PR #7, opened 2024-02-02
Updating from version 0.1.3 to 0.2

nnmkhang:
## Changelist
- Changed dependance on `symcrypttestmodule.dll` to `symcrypt.dll` 
- Updated `symcrypt` crate from `0.1.3` to `0.2`
- Updated `symcrypt-sys` crate from `0.1.0` to `0.2`
- Updated comments and `README`s to reflect change.
- Changed API signature for `symcrypt_random` from a taking a `Vec<u8>` to taking a `&mut [u8]`


PhilHallin (dismissed):
:shipit:

PhilHallin (approved):
:shipit:

README.md
```diff
@@ -10,4 +10,4 @@ Within this repository, there are 3 crates:
 
 The purpose of these crates is to bring FIPS-compliant cryptography to the Rust Ecosystem. Currently, there is only binding support for Windows and Linux.
 
-**Note:** As of version 0.1.0, only Windows AMD64 is supported.
+**Note:** As of version `0.2.0`, only `Windows AMD64`, and [`Linux Mariner`](https://github.com/microsoft/CBL-Mariner) are fully supported, with partial support for other linux distros such as `Ubuntu`.
```
mlindgren: >Linux Mariner [](http://example.com/codeflow?start=61&length=13)

np: it's now called Azure Linux #Resolved

mlindgren: >linux [](http://example.com/codeflow?start=171&length=5)

np: capitalize #Resolved

rust-symcrypt/BUILD.md
```diff
@@ -6,17 +6,19 @@ who want to use their own `symcrypt*` files.
 ## Installation
 To use the SymCrypt crate, you must have a local version of [SymCrypt](https://github.com/microsoft/SymCrypt) downloaded.
 
-Please follow the [Build Instructions](https://github.com/microsoft/SymCrypt/blob/main/BUILD.md) that is provided by SymCrypt to install SymCrypt for your target architecture.
+For ease of use, the recommended usage is to obtain these binaries from the official SymCrypt [Repo](https://github.com/microsoft/SymCrypt/releases/tag/v103.4.2).
 
-Once SymCrypt is installed and built locally on your machine, we must configure your machine so that the SymCrypt crate's build script can easily find `symcrypttestmodule.dll` and `symcrypttestmodule.lib` which are needed on Windows, or the `libsymcrypt.so*` files which are needed for Linux. 
+If you wish to build your own version please follow the [Build Instructions](https://github.com/microsoft/SymCrypt/blob/main/BUILD.md) that is provided by SymCrypt to install SymCrypt for your target architecture.
```
mlindgren: >is [](http://example.com/codeflow?start=140&length=2)

np: are #Resolved

rust-symcrypt/BUILD.md
```diff
@@ -6,17 +6,19 @@ who want to use their own `symcrypt*` files.
...
 
-Once SymCrypt is installed and built locally on your machine, we must configure your machine so that the SymCrypt crate's build script can easily find `symcrypttestmodule.dll` and `symcrypttestmodule.lib` which are needed on Windows, or the `libsymcrypt.so*` files which are needed for Linux. 
+If you wish to build your own version please follow the [Build Instructions](https://github.com/microsoft/SymCrypt/blob/main/BUILD.md) that is provided by SymCrypt to install SymCrypt for your target architecture.
+
+Once SymCrypt is installed and built locally on your machine, we must configure your machine so that the SymCrypt crate's build script can easily find `symcrypt.dll` and `symcrypt.lib` which are needed on Windows, or the `libsymcrypt.so*` files which are needed for Linux. 
 
 ### Windows Install 
 
-The `symcrypttestmodule.lib` can be found in the the following path after SymCrypt has been successfully downloaded and built. 
+The `symcrypt.lib` can be found in the the following path after SymCrypt has been successfully downloaded and built.
```
mlindgren: >downloaded and built.  [](http://example.com/codeflow?start=95&length=22)

downloaded or built? #Resolved

rust-symcrypt/BUILD.md
```diff
@@ -6,17 +6,19 @@ who want to use their own `symcrypt*` files.
...
 
 ### Windows Install 
 
-The `symcrypttestmodule.lib` can be found in the the following path after SymCrypt has been successfully downloaded and built. 
+The `symcrypt.lib` can be found in the the following path after SymCrypt has been successfully downloaded and built. 
 
 `C:\Your-Path-To-SymCrypt\SymCrypt\bin\lib`
 
-The SymCrypt crate needs a static lib to link to during its build/link time. You must configure your system so that the SymCrypt crate's build script can easily find the needed `symcrypttestmodule.lib` file.
+The SymCrypt crate needs a static lib to link to during its build/link time. You must configure your system so that the SymCrypt crate's build script can easily find the needed `symcrypt.lib` file.
```
mlindgren: This is the wrong path. If you build SymCrypt locally, there will be two distinct `symcrypt.lib` files:

1. `lib\symcrypt.lib` - this is the full static library which you would use for static linking
2. `dll\symcrypt.lib` - this is the DLL import library which is needed to dynamically link the DLL

We want to dynamically link to symcrypt.dll, so the second one is the one you want. Although, it's also possible to generate an import library directly from the DLL, and I thought bindgen supported that, so you might not really need the import library at all? #Resolved

rust-symcrypt/BUILD.md
```diff
@@ -26,18 +28,18 @@ You can configure your system one of 3 ways.
     ```
     **Note:** This change will only persist within the current process, and you must re-set the PATH environment variable after closing the PowerShell window.
 
-2. Manually copy the `symcrypttestmodule.lib` to `C:\Windows\System32`
-    Doing this will ensure that any project that uses the SymCrypt crate will be able to access `symcrypttestmodule.lib`
+2. Manually copy the `symcrypt.lib` to `C:\Windows\System32`
+    Doing this will ensure that any project that uses the SymCrypt crate will be able to access `symcrypt.lib`
```
mlindgren: I wouldn't recommend this. It's not really good practice to drop development dependencies into System32. #Resolved

rust-symcrypt/BUILD.md
```diff
@@ -26,18 +28,18 @@ You can configure your system one of 3 ways.
...
 
-3. Permanently add the lib path into your system PATH environment variable. Doing this will ensure that any project that uses the SymCrypt crate will be able to access `symcrypttestmodule.lib`
+3. Permanently add the lib path into your system PATH environment variable. Doing this will ensure that any project that uses the SymCrypt crate will be able to access `symcrypt.lib`
 
 **Option 1 or Option 2 is what is recommended for ease of use.**
 
-The `symcrypttestmodule.dll` can be found in the the following path after SymCrypt has been successfully downloaded and built. 
+The `symcrypt.dll` can be found in the the following path after SymCrypt has been successfully downloaded and built. 
 
 `C:\Your-Path-To-SymCrypt\SymCrypt\bin\exe`
```
mlindgren: This path is also wrong, it should be in the `dll` path relative to the archive root or the build output root. #Resolved

rust-symcrypt/BUILD.md
```diff
@@ -47,18 +49,18 @@ During runtime, Windows will handle finding all needed `dll`'s in order to run t
...
     ```powershell
     $env:PATH = "C:\Your-Path-To-SymCrypt\SymCrypt\bin\exe;$env:PATH"
     ```
     **Note:** This change will only persist within the current process, and you must re-set the PATH environment variable after closing the PowerShell window.
 
-3. Manually copy `symcrypttestmodule.dll` into your `C:/Windows/System32/` 
-    Doing this will ensure that any project that uses the SymCrypt crate will be able to access `symcrypttestmodule.dll`
-4. Permanently add the `symcrypttestmodule.dll` path into your System PATH environment variable. Doing this will ensure that any project that uses the SymCrypt crate will be able to access `symcrypttestmodule.lib`
+3. Manually copy `symcrypt.dll` into your `C:/Windows/System32/` 
+    Doing this will ensure that any project that uses the SymCrypt crate will be able to access `symcrypt.dll`
```
mlindgren: We probably shouldn't recommend this to external developers either, especially given that future versions of Windows will include symcrypt.dll. For third party applications, the best practice is to either rely on the in-box binary (once it's available), or ship the DLL with your application so that it will be loaded from the application directory. The latter also allows the developer to guarantee that the version of the DLL matches what their application expects. #Resolved

mlindgren:
> `~/Your-Path-To-SymCrypt/SymCrypt/bin/module/oe_full/`

This is the OpenEnclave module. On Linux, the .so files will be in the `lib` directory relative to the archive or build root.  (The terminology is a bit different on Linux, so this is different from the `lib` directory on Windows.) #Resolved

---
Refers to: rust-symcrypt/BUILD.md:70 in 6093e70. [](commit_id = 6093e70ccaf942841d7ec284d1bc29fa7d165396, deletion_comment = False)

rust-symcrypt/BUILD.md
```diff
@@ -69,5 +71,5 @@ After building SymCrypt on linux, the required `libsymcrypt.so*` files can be fo
 
 The symcrypt crate needs to be able to link with these libs during build/run time. In order to mimic the installation path for other libraries, you must place the `libsymcrypt.so*` files into the `/usr/lib/x86_64-linux-gnu/` folder. This is where the linker will look for the `libsymcrypt.so*` files. 
 
-**Note:** This process has been streamlined for `Mariner`, and there is no garuntee for combatibility on other Linux distros at the moment. 
+**Note:** This process has been streamlined for `Mariner`, and there is no guarantee for compatibility on other Linux distros at the moment.
```
mlindgren: For Azure Linux, libsymcrypt.so should be installed by default. At least, that's the case for Azure Linux 3. In 2 you might need to install it via package manager? @samuel-lee-msft might know.

More general guidance is that libsymcrypt.so needs to be placed in the linker load path. Unfortunately the way this is set varies between distros. On most distros it's set in the environment variable `$LD_LIBRARY_PATH`. On Ubuntu, it's set in this file: `/etc/ld.so.conf.d/x86_64-linux-gnu.conf`

It's still true that we don't officially support distributions other than Azure Linux and Ubuntu. #Resolved

rust-symcrypt/README.md
```diff
@@ -4,25 +4,19 @@ This crate provides friendly and idiomatic Rust wrappers over [SymCrypt](https:/
...
 This crate has a dependency on `symcrypt-sys`, which utilizes `bindgen` to create Rust/C FFI bindings.
 
-**Note:** As of version `0.1.3`, only `Windows AMD64`, and [`Linux Mariner`](https://github.com/microsoft/CBL-Mariner) are fully supported, with partial support for other linux distros such as `Ubuntu`.
-
-For ease of use, we have included a `symcrypttestmodule.dll` and `symcrypttestmodule.lib` in the `bin/amd64/` folder for users on Windows. This will only work on computers using an `AMD64 (x86_64)` architecture.
-
-We have also included the required `libsymcrypt.so` files needed for Linux users. These `.so` files have been built for the [Mariner](https://github.com/microsoft/CBL-Mariner) distro, but have been tested and confirmed working on `Ubuntu 22.04.3` via WSL. Support for other distros aside from Mariner is not guaranteed. 
-
-If you are using a different architecture, you will have to continue with the install and build steps outlined in the `BUILD.md` file.
+**Note:** As of version `0.2.0`, only `Windows AMD64`, and [`Linux Mariner`](https://github.com/microsoft/CBL-Mariner) are fully supported, with partial support for other linux distros such as `Ubuntu`.
```
mlindgren: >Linux Mariner [](http://example.com/codeflow?start=61&length=13)

Azure Linux #Resolved

rust-symcrypt/README.md
```diff
@@ -4,25 +4,19 @@ This crate provides friendly and idiomatic Rust wrappers over [SymCrypt](https:/
...
 
 ## Quick Start Guide
```
mlindgren, line 9: See comments in BUILD.md about these instructions #Resolved

rust-symcrypt/src/eckey.rs
```diff
@@ -155,9 +155,25 @@ pub(crate) fn convert_curve(curve: CurveType) -> symcrypt_sys::PCSYMCRYPT_ECURVE
 
 // get_num_format returns the correct number format needed for TLS interop since 25519 spec defines the use of Little Endian.
 pub(crate) fn get_num_format(curve_type: CurveType) -> i32 {
-    if curve_type == CurveType::Curve25519 {
-        return symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_LSB_FIRST;
-    } else {
-        return symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_MSB_FIRST;
+    // Curve25519 has only X coord, where as Nistp256 and NistP384 have X and Y coord
```
mlindgren: >// Curve25519 has only X coord, where as Nistp256 and NistP384 have X and Y coord [](http://example.com/codeflow?start=4&length=81)

This isn't relevant here, is it? #Resolved

rust-symcrypt/src/lib.rs
```diff
@@ -3,25 +3,21 @@
...
-//!
-//! For ease of use, we have included a `symcrypttestmodule.dll` and `symcrypttestmodule.lib` in the `C:\Users\<your-user>\.cargo\registry\src\github.com-****\symcrypt-0.1.3\bin\amd64` folder for users on Windows. This will only work on computers using an `AMD64 (x86_64)` architecture.
-//!
-//! We have also included the required `libsymcrypt.so` files needed for Linux users. These `.so` files have been built for the [Mariner](https://github.com/microsoft/CBL-Mariner) distro, but have been tested and confirmed working on `Ubuntu 22.04.3` via WSL. Support for other distros aside from Mariner is not guaranteed. 
-//! These files are included in the `~/.cargo/registry/src/github.com-****/symcrypt-0.1.3\bin\linux` folder for users on Linux. 
 //! 
-//! If you are using a different architecture, you will have to continue with the install and build steps outlined in the `BUILD.md` file.
-//!
+//! **Note:** As of version `0.2.0`, only `Windows AMD64`, and [`Linux Mariner`](https://github.com/microsoft/CBL-Mariner) are fully supported, with partial support for other linux distros such as `Ubuntu`.
```
mlindgren: >//! **Note:** As of version `0.2.0 [](http://example.com/codeflow?start=0&length=34)

To make this easier to maintain, I'd recommend removing all of the duplicated README content and just pointing to the README/BUILD files.

Edit: just realized while looking at the other PR that this is a documentation comment, so I guess it's expected in Rust to have something like this. Still, I think it would be good to limit this to only essential information and refer to the README for the rest. #Resolved

mlindgren (dismissed):
:clock1:

README.md
nnmkhang: Fixed

nnmkhang: fixed

rust-symcrypt/BUILD.md
nnmkhang: fix

rust-symcrypt/BUILD.md
nnmkhang: Built* fixed

rust-symcrypt/BUILD.md
nnmkhang: Im not sure about bindgen being able to generate the import library directly from the DLL, but I've just tried removing the `symcrypt.lib` from the directory and my tests fail, with error: `note: LINK : fatal error LNK1181: cannot open input file 'symcrypt.lib'`

So I think that the lib is needed. 

As for the paths I will update it to be `dll\symcrypt.lib`

I will also look into bindgen generating its own lib from the `symcrypt.dll` provided.

rust-symcrypt/src/lib.rs
nnmkhang: This is the documentation that shows up in the rust docs like this: https://docs.rs/symcrypt/0.1.3/symcrypt/

If you think that it would be better, we just point to a readme we can do that, but from what i've seen from other rust packages they have a similar layout in the `lib.rs` example rustls: https://docs.rs/rustls/0.23.4/rustls/

rust-symcrypt/BUILD.md
nnmkhang: This is what are currently recommending for users that consume our product, will follow up offline

rust-symcrypt/BUILD.md
nnmkhang: I think in the case of shipping the DLL with application makes sense, but we still need the lib to be generated and linked to.

Maybe this can be done through the Bindgen somehow. I will do some investigation. 

Will follow up offline as well

rust-symcrypt/BUILD.md
nnmkhang: > For Azure Linux, libsymcrypt.so should be installed by default. At least, that's the case for Azure Linux 3. In 2 you might need to install it via package manager? @samuel-lee-msft might know.

I spoke with @mamckee  yesterday and from our conversation it seems that Azure Linux 3 comes out the box, and for Azure Linux 2 there is a command similar to `sudo apt get` to get the `symcrypt` package 

With @mamckee's work on the Debian project do you think it might be simpler to just remove this Linux section and just say that a Debian package is coming soon?

rust-symcrypt/README.md
nnmkhang: Fixed

rust-symcrypt/README.md
nnmkhang: Following up offline

rust-symcrypt/src/eckey.rs
nnmkhang: Not really, removing.

rust-symcrypt/BUILD.md
mlindgren: Path looks good, but I'd change the first sentence here to:

"The SymCrypt crate needs to link against the SymCrypt import library during build."

mlindgren:
> `~/Your-Path-To-SymCrypt/SymCrypt/bin/module/oe_full/`

Apologies, I said something incorrect when we talked yesterday. If you download the tar.gz archive from the SymCrypt GitHub, the .so files will indeed be in the lib directory. But if you build them locally, they'll be in `<build output root>/module/generic`

I think we should recommend downloading the files from GitHub rather than building locally.

---
In reply to: [2062721955](https://github.com/microsoft/rust-symcrypt/pull/7#issuecomment-2062721955) [](http://example.com/codeflow?ancestors=2062721955)

---
Refers to: rust-symcrypt/BUILD.md:70 in 6093e70. [](commit_id = 6093e70ccaf942841d7ec284d1bc29fa7d165396, deletion_comment = False)

rust-symcrypt/README.md
mlindgren: Should be just "Azure Linux," not "Linux Azure Linux" :)

rust-symcrypt/README.md
```diff
@@ -4,27 +4,36 @@ This crate provides friendly and idiomatic Rust wrappers over [SymCrypt](https:/
...
-Copy the `symcrypttestmodule.dll` and `symcrypttestmodule.lib` from the `/bin/amd64` folder and place it into your `C:/Windows/System32` folder. 
+During runtime, Windows will handle finding all needed `dll`'s in order to run the intended program, this includes our `symcrypt.dll` file.
 
-For more information please see the `BUILD.md` file
+Here are 2 recommended options to ensure your `symcrypt.dll` is found by Windows during runtime, for more info please see [Dynamic-link library search order](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order).
+
+1. Put the `symcrypt.dll` in the same folder as your output `.exe` file. If you are doing development (not release), the common path will be: `C:\your-project\target\debug\`.
+2. Permanently add the `symcrypt.dll` path into your System PATH environment variable. Doing this will ensure that any project that uses the SymCrypt crate will be able to access `symcrypt.lib`
+
+**Note this is a temporary solution while we work on shipping SymCrypt out of the box on Windows.**
```
mlindgren: This isn't really private but I don't know if we should mention it here just because we don't really have a timeline/release vehicle for it yet. #Resolved

rust-symcrypt/BUILD.md
mlindgren: Current wording looks good to me, but maybe we can add instructions for Azure Linux?

For Azure Linux 2.0 it should just be `dnf install symcrypt` (note: I haven't tested this myself)

In 3.0 it will be included in-box, but 3.0 hasn't shipped yet

rust-symcrypt/README.md
```diff
@@ -4,27 +4,36 @@ This crate provides friendly and idiomatic Rust wrappers over [SymCrypt](https:/
...
+
+For more information please see the `BUILD.md` file on the [`rust-symcrypt`](https://github.com/microsoft/rust-symcrypt/tree/main/rust-symcrypt) page
 
 ### Linux:
-Copy all of the `libsymcrypt.so*` files from the `/bin/linux` folder and place it into your `/usr/bin/x86_64-linux-gnu/` folder. 
+Download and all of the `libsymcrypt.so*` files for you corresponding CPU architecture from the [SymCrypt Releases Page](https://github.com/microsoft/SymCrypt/releases/tag/v103.4.2).
+
+Support for `Debian` and `Ubuntu` via package manager is in the works, for now you must place the `libsymcrypt.so*` files into linker load path. The way that this is set will vary between distros. On most distros it set via the environment variable `$LD_LIBRARY_PATH`.
```
mlindgren, line 30: >Support for `Debian` and `Ubuntu` via package manager is in the works [](http://example.com/codeflow?start=0&length=69)

Maybe mention here as well that it is already available on Azure Linux via package manager #Resolved

rust-symcrypt/src/lib.rs
mlindgren: This looks good to me. Same comment about "Linux Azure Linux" though

mlindgren (changes requested):
New changes look good, we're almost there. Just requested a few more small tweaks

README.md
mlindgren: Not fixed?

mlindgren: I think you forgot to commit the updates to this file

rust-symcrypt/README.md
nnmkhang: Good point, I will remove

rust-symcrypt/README.md
nnmkhang: Will note that.

README.md
nnmkhang: Fixed now.

rust-symcrypt/src/lib.rs
nnmkhang: Fixed

rust-symcrypt/README.md
nnmkhang: Fixed

rust-symcrypt/BUILD.md
nnmkhang: I've updated the `BUILD.MD` to have an install section for Azure Linux, and I've tried the `tdnf install symcrypt` before on my machine and it works. I've also included a note about Azure Linux 3.0

rust-symcrypt/BUILD.md
nnmkhang: Updated the line to match your suggestion

README.md
nnmkhang: Fixed now

mlindgren (approved):
:shipit:

PR #16, opened 2024-04-24
Adding support for SHA1, MD5 and SHA512

nnmkhang:
Overview:
- Adding Hashing support for SHA1, MD5 and SHA512
- Adding HMAC support for SHA1, MD5 and SHA512
- Added Unit tests for SHA1, MD5 and SHA512
- Adding binding support for future algos: RSA, PSS, DSA
- Updating comments
- Locking SHA1 and MD5 behind feature flag

Notes: 
- This will be a github only contribution, we will update the crate with a major release once the rest of the algorithms are complete.


rust-symcrypt/src/hash.rs
```diff
@@ -244,8 +449,94 @@ pub fn sha384(data: &[u8]) -> [u8; SHA384_RESULT_SIZE] {
...
+            )
+        }
+    }
+}
+
+/// Stateless hash function for SHA512.
+///
+/// `data` is a reference to an array of arbitrary length.
+///
+/// `result` is an array of size `SHA512_RESULT_SIZE`, which is 60 bytes. This call cannot fail.
```
PhilHallin: >60 [](http://example.com/codeflow?start=64&length=2)

64 #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -59,14 +77,22 @@ use std::ptr;
...
+#[cfg(feature = "md5")]
+pub const MD5_HMAC_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_MD5_RESULT_SIZE as usize;
+/// 20
+#[cfg(feature = "sha1")]
+pub const SHA1_HMAC_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_SHA1_RESULT_SIZE as usize;
 /// 32
 pub const SHA256_HMAC_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_SHA256_RESULT_SIZE as usize;
 /// 48
 pub const SHA384_HMAC_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_SHA384_RESULT_SIZE as usize;
+/// 60
```
PhilHallin: >60 [](http://example.com/codeflow?start=4&length=2)

64 #Closed

symcrypt-bindgen/build.rs
```diff
@@ -41,6 +41,15 @@ fn main() {
         .allowlist_function("^(SymCryptEckey(Allocate|Free|SizeofPublicKey|GetValue|SetRandom|SetValue|SetRandom|))$")
         .allowlist_var("SYMCRYPT_FLAG_ECKEY_ECDH")
         .allowlist_function("SymCryptEcDhSecretAgreement")
+        // RSA FUNCTIONS
+        .allowlist_function("^(SymCryptRsakey.*)$")
+        .allowlist_function("^(SymCryptRsaRaw.*)$")
+        // DSA FUNCTIONS
+        .allowlist_function("^(SymCryptDsa(Sign|Verify).*)$")
```
PhilHallin: Why Dsa instead of EcDsa, EcDh #Closed

rust-symcrypt/src/hash.rs
nnmkhang: Fixed

rust-symcrypt/src/hmac.rs
nnmkhang: Fixed

symcrypt-bindgen/build.rs
nnmkhang: EcDh is already added from the previous work , Will add in EcDsa and remove DSA.

nnmkhang:
@mlindgren Code is more or less copy paste of `SHA256` and `SHA384`. This mirrors what SymCrypt has at the moment, I think doing it like this will provided good parity. I know that there is also a generic `SHAX` function that lets you pass in the algorithm, but I think that it would be a very good onboarding ticket for a new hire so my opinion is that we save it.

rust-symcrypt/src/hash.rs
```diff
@@ -1,5 +1,24 @@
 //! Hashing functions. For further documentation please refer to symcrypt.h
 //!
+//!
+//! # Supported Hashing functions
```
mlindgren, line 4: General comment about this file - can't remember if I've mentioned this before.

SymCrypt has generic hash functions, `SymCryptHash`, `SymCryptHashInit`, `SymCryptHashAppend`, and `SymCryptHashResult` which you can use with any supported hashing function. You just pass in the hash function to use as an argument. You could probably substantially reduce the amount of duplicated boilerplate code in this file by leveraging those functions. I'm not sure exactly how best to do that in Rust, but it would probably be something like having an additional trait parameter for the hash type and then mapping that to a `PCSYMCRYPT_HASH` function which you could pass into SymCrypt's generic hash functions.

Edit: just noticed your other comment, will send you a message.

rust-symcrypt/src/hmac.rs
```diff
@@ -1,5 +1,22 @@
 //! Hmac functions. For further documentation please refer to symcrypt.h
```
mlindgren, line 1: Likewise for HMAC - you can use SymCrypt's generic HMAC functions here.

rust-symcrypt/src/hmac.rs
```diff
@@ -84,6 +110,372 @@ pub trait HmacState: Clone {
     fn result(self) -> Self::Result;
 }
 
+/// `HmacMd5ExpandedKey` is a struct that represents the expanded key for the [`HmacMd5State`].
+#[cfg(feature = "md5")]
+struct HmacMd5ExpandedKey(symcrypt_sys::SYMCRYPT_HMAC_MD5_EXPANDED_KEY);
+// Wrapping the expanded key so that it can be independently dropped after it's ref count has gone to 0
+
+// Since HmacMd5ExpandedKey can be referenced multiple times, HmacMd5ExpandedKey must be ref counted and a there needs to be a separate drop()
```
mlindgren: > a [](http://example.com/codeflow?start=104&length=2)

nitpick: extra word

rust-symcrypt/src/hmac.rs
```diff
@@ -84,6 +110,372 @@ pub trait HmacState: Clone {
...
+
+#[cfg(feature = "md5")]
+struct HmacMd5Inner {
+    state: symcrypt_sys::SYMCRYPT_HMAC_MD5_STATE,
+
+    // Must Arc<> the expanded_key field since it must be ref counted, clones of HmacSha265State will reference the same expanded key. Pin<> on the key
+    // is to ensure that address is not moved throughout the expanded key's lifetime.
+    //
+    // This semantic is not needed for the state field since it is initialized in line with HmacMd5Inner initialization.
+    expanded_key: Pin<Arc<HmacMd5ExpandedKey>>,
```
mlindgren, line 152: >Pin<Arc [](http://example.com/codeflow?start=18&length=7)

Just curious, does it matter which order these go in? I would have thought `Arc<Pin<...>>` would make more sense but it might not matter.

rust-symcrypt/src/hmac.rs
```diff
@@ -84,6 +110,372 @@ pub trait HmacState: Clone {
...
+}
+
+#[cfg(feature = "md5")]
+unsafe impl Send for HmacMd5Inner {
+    // Md5 Send/Sync is not needed for rustls, adding for consistency.
```
mlindgren: >// Md5 Send/Sync is not needed for rustls, adding for consistency. [](http://example.com/codeflow?start=4&length=66)

It doesn't really make sense to use a single hash object across multiple threads concurrently. By implementing this, doesn't it imply that this struct is thread-safe? But SYMCRYPT_HMAC_MD5_STATE is *not* thread safe.

mlindgren (changes requested):
:clock1:

rust-symcrypt/src/hash.rs
mlindgren: Resolved offline

rust-symcrypt/src/hmac.rs
mlindgren: Resolved offline

rust-symcrypt/src/hmac.rs
nnmkhang: Yeah it does matter, For `Pin<>` it pins the memory location, and for `Arc<>` it makes it shareable and ref counted between threads. With `Arc<>` there is some extra buffers and padding that goes around that would not be pinned if it were instead `Arc<Pin<>>`

rust-symcrypt/src/hmac.rs
nnmkhang: Spoke offline. New github issue filed, and we will remove the `Send/Sync` traits for algos not required by rustls

rust-symcrypt/src/hmac.rs
nnmkhang: Fixed for this one and the other hmacing algos.

rust-symcrypt/src/hmac.rs
nnmkhang: Removing these Send/Sync functions for the ones not needed for rustls, filed github issue.

mlindgren (approved):
:shipit:

PR #22, opened 2024-06-28
Rename md5 and sha1 to weak variants

asaintpi:
Updates made:

Updated name from sha1 to weak-sha1 in hash.rs and hmac.rs
Updated name from md5 to weak-md5 in hash.rs and hmac.rs
Updated documentation around weak-sha1 and weak-md5
Updated Cargo.toml to reflect.

### DIFF
--- a/rust-symcrypt/Cargo.toml
+++ b/rust-symcrypt/Cargo.toml
@@ -22,7 +22,7 @@ lazy_static = "1.4.0"
 default = []
 md5 = []
 sha1 = []
-weak-crypto = ["md5","sha1"]
+weak-crypto = ["weak-md5","weak-sha1"]
 
 [dev-dependencies]
 hex = "0.4.3"
--- a/rust-symcrypt/src/hash.rs
+++ b/rust-symcrypt/src/hash.rs
@@ -3,15 +3,15 @@
 //!
 //! # Supported Hashing functions
 //! ```ignore
-//! Md5 // Note: Md5 is disabled by default, to enable pass the md5 flag
-//! Sha1 // Note: Sha1 is disabled by default, to enable pass the sha1 flag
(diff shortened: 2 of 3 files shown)


rust-symcrypt/Cargo.toml
```diff
@@ -22,7 +22,7 @@ lazy_static = "1.4.0"
 default = []
 md5 = []
 sha1 = []
```
nnmkhang, line 24: these should be `weak-sha1` and `weak-md5`. You were correct with changing the `weak-crypto` feature, but right now when you apply `weak-crypto` you are enabling `md5` and `sha1` which is not implemented anywhere in the code

rust-symcrypt/src/hash.rs
```diff
@@ -535,7 +535,7 @@ pub fn sha512(data: &[u8]) -> [u8; SHA512_RESULT_SIZE] {
 
 #[cfg(test)]
 mod test {
-    // Note: by default sha1 and md5 are turned off, to enable for testing you can use:
+    // Note: by default "weak-sha1" and "weak-md5" are turned off, to enable for testing you can use:
     // cargo test --features "weak-crypto"
```
nnmkhang, line 539: did you try running this command and verifying that the tests are being called after your changes?

nnmkhang:
Please update the documentation on `README.md` and `lib.rs` to communicate the new naming convention

nnmkhang (changes requested):
LGTM,

Left a few comments. 

Double check the Cargo.toml as well as run the test with the flag enabled to see that the flags are being called properly 

also nitpick: You should name your branches in a way that is tied to the user more clearly. ie: user/asauntpi/weak-md5-sha1-names

mmwuyan (approved):
aside from Khang's comments, lgtm

PR #29, opened 2024-08-05
User/nnmkhang/hashing pin

### LINKED ISSUE #27: Add phantom unpin for Pin::Box'd code
PhantomPinned in std::marker - Rust (rust-lang.org)](https://doc.rust-lang.org/std/marker/struct.PhantomPinned.html)

The reason is that the &mut *instance.0 line is likely using this DerefMut impl: https://doc.rust-lang.org/std/pin/struct.Pin.html#impl-DerefMut-for-Pin%3CPtr%3E
 
That's only implemented for types that are Unpin, but that's an auto-trait, so most things are Unpin unless you opt out.

Generally getting a mutable reference through a Pin has to be unsafe because otherwise you can use mem::swap to move out of the reference.
 
Example: Sha256State.

Our code right now does not move the Sha256State after creation and deletion, but if somehow it is moved after the Sha256State has been created it would `Unpin` and cause an error. To avoid this you should look into PhantomPinning

nnmkhang:
States that use `Pin<T>` to ensure that states stay consistent throughout their lifetime must also include `PhantomPin` to ensure that the memory does not get moved by rust compiler optimizations.

### DIFF
--- a/rust-symcrypt/src/gcm.rs
+++ b/rust-symcrypt/src/gcm.rs
@@ -63,8 +63,12 @@
 //!
 use crate::block_ciphers::*;
 use crate::errors::SymCryptError;
+use std::marker::PhantomPinned;
 use std::pin::Pin;
 use symcrypt_sys;
+use core::ffi::c_void;
+use std::mem;
+use std::ptr;
 
 /// [`GcmExpandedKey`] is a struct that holds the Gcm expanded key from SymCrypt.
 pub struct GcmExpandedKey {
@@ -75,8 +79,53 @@ pub struct GcmExpandedKey {
 
     // SymCrypt expects the address for its structs to stay static through the structs lifetime to guarantee that structs are not memcpy'd as
     // doing so would lead to use-after-free and inconsistent states.
-    expanded_key: Pin<Box<symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY>>,
+    expanded_key: Pin<Box<GcmInnerKey>>,
     key_length: usize,
+}
+
+struct GcmInnerKey {
+    // inner represents the actual state of the hash from SymCrypt
+    inner: symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY,
+
+    // _pinned is a marker to ensure that instances of the inner state cannot be moved once pinned.
+    // This prevents the struct from implementing the Unpin trait, enforcing that any
+    // references to this structure remain valid throughout its lifetime.
+    _pinned: PhantomPinned,
+}
+
+impl Drop for GcmInnerKey {
+    fn drop(&mut self) {
+        unsafe {
+            // SAFETY: FFI calls
+            symcrypt_sys::SymCryptWipe(
+                ptr::addr_of_mut!(self.inner) as *mut c_void, // Using addr_of_mut! so we don't access in the inner field
+                mem::size_of_val(&self.inner) as symcrypt_sys::SIZE_T, // Using size_of_val! so we don't access in the inner field
+            );
+        }
+    }
+}
+
+impl GcmInnerKey {
+    /// Creates a new GcmInnerKey and returns a pinned Box<Self>
+    fn new() -> Pin<Box<Self>> {
+        Box::pin(GcmInnerKey {
+            inner: symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY::default(),
+            _pinned: PhantomPinned,
+        })
+    }
+
+    /// Provides a mutable pointer to the inner SymCrypt state.
+    ///
+    /// This is primarily meant to be used while making calls to the underlying SymCrypt APIs.
+    fn get_inner_mut(self: Pin<&mut Self>) -> *mut symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY {
+        // SAFETY: We are returning a pointer to the pinned data, and Rust guarantees that this data cannot be moved.
+        unsafe { &mut self.get_unchecked_mut().inner as *mut _ }
+    }
+
+    // Safe method to access the inner state immutably
+    pub(crate) fn get_inner(&self) -> *const symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY {
+        &self.inner as *const _
+    }
 }
 
 /// `encrypt_in_place` and `decrypt_in_place` take in an allocated `buffer` as an in/out parameter for performance reasons.
@@ -86,8 +135,10 @@ impl GcmExpandedKey {
     /// This function can fail and will propagate the error back to the caller. This call will fail if the wrong key size is provided.
     /// The only accepted Cipher for GCM is [`BlockCipherType::AesBlock`]
     pub fn new(key: &[u8], cipher: BlockCipherType) -> Result<Self, SymCryptError> {
-        let mut expanded_key = Box::pin(symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY::default()); // boxing here so that the memory is not moved
-        gcm_expand_key(key, &mut expanded_key, convert_cipher(cipher))?;
+        let mut expanded_key = GcmInnerKey::new(); // Get expanded_key that is already Pin<Box<T>>'d
+
+        // Use as_mut() to get a Pin<&mut GcmInnerKey> and then call get_inner_mut to get *mut 
+        gcm_expand_key(key,  expanded_key.as_mut().get_inner_mut(), convert_cipher(cipher))?;
         let gcm_expanded_key = GcmExpandedKey {
             expanded_key: expanded_key,
             key_length: key.len(),
@@ -108,7 +159,7 @@ impl GcmExpandedKey {
         unsafe {
             // SAFETY: FFI calls
             symcrypt_sys::SymCryptGcmEncrypt(
-                &*self.expanded_key,
+                self.expanded_key.get_inner(),
                 nonce.as_ptr(),
                 nonce.len() as symcrypt_sys::SIZE_T,
                 auth_data.as_ptr(),
@@ -136,7 +187,7 @@ impl GcmExpandedKey {
         unsafe {
             // SAFETY: FFI calls
             match symcrypt_sys::SymCryptGcmDecrypt(
-                &*self.expanded_key,
+                self.expanded_key.get_inner(),
                 nonce.as_ptr(),
                 nonce.len() as symcrypt_sys::SIZE_T,
                 auth_data.as_ptr(),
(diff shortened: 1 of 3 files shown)


rust-symcrypt/src/hash.rs
```diff
@@ -372,7 +373,16 @@ pub fn sha1(data: &[u8]) -> [u8; SHA1_RESULT_SIZE] {
 }
 
 /// [`Sha256State`] is a struct that represents a stateful sha256 hash and implements the [`HashState`] trait.
-pub struct Sha256State(Pin<Box<symcrypt_sys::SYMCRYPT_SHA256_STATE>>);
+pub struct Sha256State(Pin<Box<Sha256InnerState>>);
+pub struct Sha256InnerState {
```
eholk: I like what you did here. I feel like I was telling you to reach into the internals of `symcrypt_sys::SYMCRYPT_SHA256_STATE` and add the `PhantomPinned` field there, but if I remember right, that's generated by bindgen so if you had manual edits that would make maintenance harder. This way accomplishes the same thing but it more maintainable.

rust-symcrypt/src/hash.rs
```diff
@@ -381,13 +391,26 @@ pub struct Sha256State(Pin<Box<symcrypt_sys::SYMCRYPT_SHA256_STATE>>);
...
         instance
     }
+
+    // Safe method to access the inner state mutably
+    pub(crate) fn get_inner_mut(&mut self) -> *mut symcrypt_sys::SYMCRYPT_SHA256_STATE {
+        unsafe { 
+            // SAFETY: getting a mutable raw pointer to the inner state
+            &mut self.0.as_mut().get_unchecked_mut().inner as *mut _
+        }
+    }
```
eholk: This function makes me a little nervous, but I think it's technically correct.

What saves you is that you are returning a `*mut` instead of an `&mut`. If this returned `&mut`, you'd be able to use it to get around the restrictions `Pin` tries to impose. But, since you return `*mut`, dereferencing that pointer is unsafe.

I would propose a couple of things to make this more robust.

First, we need to document the fact that the `*mut` that's returned is also pinned, even though that's not carried in the type. This means users must be careful not to move out of that pointer. As I mentioned before, doing so would require an `unsafe` block, but it's still good practice to document what requirements we're placing on unsafe code. I'd suggest something like:

```suggestion
    /// Get a mutable pointer to the inner SymCrypt state
    ///
    /// This is primarily meant to be used while making calls to the underlying SymCrypt APIs.
    ///
    /// ## Safety:
    /// 
    /// This function returns pointer to pinned data, which means callers must not use the pointer to move the data out of its location.
    fn get_inner_mut(&mut self) -> *mut symcrypt_sys::SYMCRYPT_SHA256_STATE {
        unsafe { 
            // SAFETY: getting a mutable raw pointer to the inner state
            &mut self.0.as_mut().get_unchecked_mut().inner as *mut _
        }
    }
```

Second, is it possible to remove the `pub(crate)` part of this? Keeping this function as tightly scoped as possible can help limit the opportunities to misuse it. I made that change in my diff above because it looks like right now this function is only used in this file.

rust-symcrypt/src/hash.rs
```diff
@@ -381,13 +391,26 @@ pub struct Sha256State(Pin<Box<symcrypt_sys::SYMCRYPT_SHA256_STATE>>);
...
+    
+    // Safe method to access the inner state immutably
+    pub(crate) fn get_inner(&self) -> *const symcrypt_sys::SYMCRYPT_SHA256_STATE {
+        &self.0.as_ref().get_ref().inner as *const _
+    }
```
eholk, line 479: Incidentally, this function does not make me as nervous because a const pointer or reference cannot be used to move a value. That also shows up in the fact that you did need to use any `unsafe` in its body.

eholk (approved):
This is pretty much right!

I had some suggestions around the `get_inner_mut` function, but that was mainly around documenting how to use it correctly.

rust-symcrypt/src/hash.rs
nnmkhang: Yeah the manual edits was the bit reason I moved it over to an inner state

rust-symcrypt/src/gcm.rs
```diff
@@ -75,19 +79,66 @@ pub struct GcmExpandedKey {
...
+    fn new() -> Pin<Box<Self>> {
+        Box::pin(GcmInnerKey {
+            inner: symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY::default(),
+            _pinned: PhantomPinned,
+        })
+    }
+
+    /// Provides a mutable pointer to the inner SymCrypt state.
+    ///
+    /// This is primarily meant to be used while making calls to the underlying SymCrypt APIs.
```
eholk, line 108: It'd be good to add a comment here stating that the pointer returned is pinned so callers need to be careful not to move it.

rust-symcrypt/src/hash.rs
```diff
@@ -195,13 +207,28 @@ pub struct Md5State(Pin<Box<symcrypt_sys::SYMCRYPT_MD5_STATE>>);
...
         }
         instance
     }
+
+    /// Get a mutable pointer to the inner SymCrypt state
+    ///
+    /// This is primarily meant to be used while making calls to the underlying SymCrypt APIs.
+    fn get_inner_mut(&mut self) -> *mut symcrypt_sys::SYMCRYPT_MD5_STATE {
+        unsafe { 
+            // This function returns pointer to pinned data, which means callers must not use the pointer to move the data out of its location.
```
eholk: It'd be good to state the part about what callers need to do in the doc comment so that the callers can see it too :)

rust-symcrypt/src/hmac.rs
```diff
@@ -481,16 +515,38 @@ pub struct HmacSha256State {
     // Using an `HmacSha256Inner` state that is Pin<Box<>>'d. Memory allocation is not handled by SymCrypt and Self is moved
     // around when returning from `HmacSha256State::new()`. Box<> heap allocates the memory and ensures that it does not move
     // within its lifetime.
-    inner: Pin<Box<HmacSha256Inner>>,
+    state: Pin<Box<HmacSha256Inner>>,
+
+    // Must Arc<> the expanded_key field since it must be ref counted, clones of HmacSha265State will reference the same expanded key.
+    // Arc<T> pointer can move, but its reference T is Pin<Box<>>'d to ensure that the address does not move during its lifetime.
+    key: Arc<Pin<Box<HmacSha256ExpandedKey>>>,
```
eholk, line 634: I think a `Pin<Arc<HmacSha256ExpandedKey>>` would likely work here.

One snag you might run into is that you can't generally mutate the inside of an `Arc` without a `Mutex`. But, if you have the pattern of needing to initialize it before you share it and then it becomes immutable, you can use [`Arc::get_mut()`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut).

rust-symcrypt/src/hash.rs
eholk: This applies to the similar methods further down too.

rust-symcrypt/src/hmac.rs
nnmkhang: I don't see any reason that they would ever need to modify the inside of the `Arc` in this case. The arc is used to ref count the usage of the key so I think it should be fine like this. I will keep the `Arc::get_mut()` in mind for the future

rust-symcrypt/src/hash.rs
nnmkhang: I'll update it to be more clear with doc comments as well

rust-symcrypt/src/gcm.rs
nnmkhang: Sounds good, I will do a comment overhaul in general but will make sure to address this

rust-symcrypt/src/hash.rs
```diff
@@ -251,9 +282,9 @@ impl Drop for Md5State {
         unsafe {
             // SAFETY: FFI calls
             symcrypt_sys::SymCryptWipe(
-                ptr::addr_of_mut!(self.0) as *mut c_void,
-                mem::size_of_val(&mut self.0) as symcrypt_sys::SIZE_T,
-            )
+                self.get_inner_mut() as *mut c_void,
+                mem::size_of_val(&*self.get_inner()) as symcrypt_sys::SIZE_T,
```
PhilHallin, line 286: Did you debug breakpoint here to confirm? #Closed

rust-symcrypt/src/gcm.rs
```diff
@@ -75,19 +79,73 @@ pub struct GcmExpandedKey {
...
+    // This prevents the struct from implementing the Unpin trait, enforcing that any
+    // references to this structure remain valid throughout its lifetime.
+    _pinned: PhantomPinned,
+}
+
+impl Drop for GcmInnerKey {
+    fn drop(&mut self) {
+        unsafe {
+            // SAFETY: FFI calls
+            symcrypt_sys::SymCryptWipe(
```
PhilHallin: >symcrypt_sys::SymCryptWipe( [](http://example.com/codeflow?start=12&length=27)

Why different from hash drop

                self.get_inner_mut() as *mut c_void,
                mem::size_of_val(&*self.get_inner()) as symcrypt_sys::SIZE_T,
            );

Shouldn't you call the get_inner functions? #Closed

PhilHallin:
> #[cfg(feature = "md5")]

Update MD5? #Closed

---
Refers to: rust-symcrypt/src/hmac.rs:115 in 8b5da01. [](commit_id = 8b5da019c9380cfa485ab74a9eef9ea6cdad517b, deletion_comment = False)

PhilHallin:
> #[cfg(feature = "sha1")]

Update SHA1? #Closed

---
Refers to: rust-symcrypt/src/hmac.rs:288 in 8b5da01. [](commit_id = 8b5da019c9380cfa485ab74a9eef9ea6cdad517b, deletion_comment = False)

rust-symcrypt/src/gcm.rs
nnmkhang: It has to do with how the structs are set up, I am dropping `GcmInnerKey` which is the <Pin<Box>> itself, where as `Sha256State` drops the state which holds the `Sha256Inner`, 

For the case of Gcm, it does not matter since the key wont be re-used, I will change over to the pattern that I have for Sha and Hmac.

rust-symcrypt/src/hash.rs
nnmkhang: Confirmed with debugger, also confirmed ref counting for Hmac Expanded Key

rust-symcrypt/src/gcm.rs
```diff
@@ -75,19 +78,73 @@ pub struct GcmExpandedKey {
 
     // SymCrypt expects the address for its structs to stay static through the structs lifetime to guarantee that structs are not memcpy'd as
     // doing so would lead to use-after-free and inconsistent states.
-    expanded_key: Pin<Box<symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY>>,
+    expanded_key: Pin<Box<GcmInnerKey>>,
     key_length: usize,
 }
 
+impl Drop for GcmExpandedKey {
```
PhilHallin: >impl Drop for GcmExpandedKey  [](http://example.com/codeflow?start=0&length=29)

Seems like this drop should be on the GcmInnerKey #Closed

PhilHallin (approved):
:shipit:

PR #30, opened 2024-08-09
Adding support for ECC, making breaking changes for ECDH

nnmkhang:
Additions:
- Added code for Ecdsa
- Made centralized location for eckey via mod.rs
- Added supporting documentation

Deletions:
- Removed ecdh.rs

Breaking Changes:
- Breaking change for ECDH, all callers must update their semantics for ECDH calls.


rust-symcrypt/src/ecc/mod.rs
```diff
@@ -0,0 +1,623 @@
...
+// EcKey is a wrapper around symcrypt_sys::PSYMCRYPT_ECKEY.
+pub(crate) struct InnerEcKey(symcrypt_sys::PSYMCRYPT_ECKEY);
+
+// Must drop the EcKey before the expanded EcCurve is dropped
+// EcCurve has static lifetime so this will always be the case.
+impl Drop for InnerEcKey {
+    fn drop(&mut self) {
+        unsafe {
+            // SAFETY: FFI calls
+            symcrypt_sys::SymCryptEckeyFree(self.0);
```
PhilHallin: >SymCryptEckeyFree [](http://example.com/codeflow?start=26&length=17)

I don't believe SymCryptEckeyFree checks for NULL pointer. You have a case where the SymCryptEckeyAllocate can fail. #Closed

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -0,0 +1,623 @@
...
+unsafe impl Sync for InnerEcCurve {
+    // TODO: Discuss send/sync for rustls
+}
+
+// Must drop EcCurve after EcKey is dropped, will always be the case since EcCurve is static.
+impl Drop for InnerEcCurve {
+    fn drop(&mut self) {
+        unsafe {
+            // SAFETY: FFI calls
+            symcrypt_sys::SymCryptEcurveFree(self.0)
```
PhilHallin: >self.0 [](http://example.com/codeflow?start=45&length=6)

Might also want to check if nonNULL #Closed

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -0,0 +1,623 @@
...
+    pub fn generate_key_pair(
+        curve_type: CurveType,
+        ec_key_usage: EcKeyUsage,
+    ) -> Result<Self, SymCryptError> {
+        let ec_curve = InnerEcCurve::new(curve_type); // Can fail here due to insufficient memory.
+        unsafe {
+            // SAFETY: FFI calls
+            // Stack allocated since we will do SymCryptEckeyAllocate.
+            let key_ptr = InnerEcKey(symcrypt_sys::SymCryptEckeyAllocate(ec_curve.0));
+            if key_ptr.0.is_null() {
```
PhilHallin, line 184: > if key_ptr.0.is_null() { [](http://example.com/codeflow?start=11&length=25)

The drop will be called here where the ptr is NULL. #Closed

rust-symcrypt/src/ecc/mod.rs
PhilHallin: SymCryptEcurveFree() doesn't check for null

PhilHallin (approved):
:shipit:

rust-symcrypt/src/lib.rs
```diff
@@ -82,8 +83,7 @@ use std::sync::Once;
 
 pub mod block_ciphers;
 pub mod chacha;
-pub mod ecdh;
-pub mod eckey;
+pub mod ecc;
```
mlindgren, line 86: How do you handle versioning for this library? Changing the name of this module would be a breaking change, right? #Resolved

rust-symcrypt/src/ecc/ecdh.rs
```diff
@@ -0,0 +1,253 @@
...
+//! ```
+//!
+use crate::ecc::{curve_to_num_format, EcKey};
+use crate::errors::SymCryptError;
+use std::vec;
+use symcrypt_sys;
+
+/// Impl for EcDh struct.
+impl EcKey {
+    /// `ecdh_secret_agreement()` returns a [`EcDhSecretAgreement`] that represents the secret agreement between the private key and the public key,
```
mlindgren: >returns a [`EcDhSecretAgreement`] [](http://example.com/codeflow?start=34&length=33)

It returns a `Vec<u8>` which is the agreed secret, right? #Resolved

mlindgren: >that represents the secret agreement between the private key and the public key [](http://example.com/codeflow?start=68&length=79)

This phrasing is kind of weird; I'd probably just say that it returns the agreed secret as a `Vec<u8>`

rust-symcrypt/src/ecc/ecdh.rs
```diff
@@ -0,0 +1,253 @@
...
+    ///  or a [`SymCryptError`] if the operation failed.
+    ///
+    /// `public_key` is an [`EcKey`] that represents the public key that the secret agreement is being calculated with.
+    ///
+    /// If the key usage is not [`EcKeyUsage::EcDh`], the function will return a [`SymCryptError::InvalidArgument`].
```
mlindgren: >is not [`EcKeyUsage::EcDh`] [](http://example.com/codeflow?start=25&length=27)

I can't find the definition of `EcKeyUsage`...

Also, I don't remember where we landed on how to represent dual-use keys. So this might need to say `EcKeyUsage::EcDh` or `EcKeyUsage::Both`, if the latter is an option.

Edit: just realized it's new in this PR, which makes sense in retrospect. So it should include `EcKeyUsage::EcDhAndEcDsa`

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+    /// `curve_type` represents the [`CurveType`] of the key.
+    ///
+    /// `ec_key_usage` represents the [`EcKeyUsage`] of the key and indicates if the key will be used for [`EcKeyUsage::EcDsa`] or [`EcKeyUsage::EcDh`], or [`EcKeyUsage::EcDhAndEcDsa`].
+    ///
+    /// This function will return a [`SymCryptError::MemoryAllocationFailure`] if there is not enough memory to allocate the key.
+    pub fn generate_key_pair(
+        curve_type: CurveType,
+        ec_key_usage: EcKeyUsage,
+    ) -> Result<Self, SymCryptError> {
+        let ec_curve = InnerEcCurve::new(curve_type); // Can fail here due to insufficient memory.
```
mlindgren: >// Can fail here due to insufficient memory. [](http://example.com/codeflow?start=54&length=44)

Need to check this for null before calling `SymCryptEckeyAllocate`, don't you?

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+    ///
+    /// `ec_key_usage` represents the [`EcKeyUsage`] of the key and indicates if the key will be used for [`EcKeyUsage::EcDsa`] or [`EcKeyUsage::EcDh`], or [`EcKeyUsage::EcDhAndEcDsa`].
+    ///
+    ///  This function will return a [`SymCryptError::MemoryAllocationFailure`] if there is not enough memory to allocate the key.
+    pub fn set_public_key(
+        curve_type: CurveType,
+        public_key: &[u8],
+        ec_key_usage: EcKeyUsage,
+    ) -> Result<Self, SymCryptError> {
+        let ec_curve = InnerEcCurve::new(curve_type); // Can fail here due to insufficient memory.
```
mlindgren: >// Can fail here due to insufficient memory. [](http://example.com/codeflow?start=54&length=44)

Same here

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+    /// it must match the format of the private key and will fail with [`SymCryptError::InvalidArgument`] if it does not.
+    ///
+    /// `ec_key_usage` represents the [`EcKeyUsage`] of the key and indicates if the key will be used for [`EcKeyUsage::EcDsa`] or [`EcKeyUsage::EcDh`], or [`EcKeyUsage::EcDhAndEcDsa`].
+    pub fn set_key_pair(
+        curve_type: CurveType,
+        private_key: &[u8],
+        public_key: Option<&[u8]>,
+        ec_key_usage: EcKeyUsage,
+    ) -> Result<Self, SymCryptError> {
+        let ec_curve = InnerEcCurve::new(curve_type); // Can fail here due to insufficient memory.
```
mlindgren: >// Can fail here due to insufficient memory. [](http://example.com/codeflow?start=54&length=44)

Same here

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+        unsafe {
+            //SAFETY: FFI calls
+            // Stack allocated since we will do SymCryptEckeyAllocate.
+            let key_ptr = InnerEcKey(symcrypt_sys::SymCryptEckeyAllocate(ec_curve.0));
+
+            if key_ptr.0.is_null() {
+                return Err(SymCryptError::MemoryAllocationFailure);
+            }
+
+            let (public_key_ptr, public_key_count) = match public_key {
```
mlindgren: >public_key_count [](http://example.com/codeflow?start=33&length=16)

`public_key_length`? Count makes it sound like it's the number of public keys #Resolved

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+                    };
+                    Ok(key)
+                }
+                err => Err(err.into()),
+            }
+        }
+    }
+
+    /// `export_public_key()` returns the public key associated with the [`EcKey`] as a `Vec<u8>`.
+    pub fn export_public_key(&self) -> Result<Vec<u8>, SymCryptError> {
```
mlindgren, line 307: >export_public_key [](http://example.com/codeflow?start=11&length=17)

We don't allow export of the private key? #Resolved

rust-symcrypt/src/ecc/ecdsa.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// `hashed_message` is a `&[u8]` that represents the hash value to sign.
+    ///
+    /// If the key usage is not [`EcKeyUsage::EcDsa`], or [`EcKeyUsage::EcDhAndEcDsa`] the function will
+    /// fail with a [`SymCryptError`] with the value [`SymCryptError::InvalidArgument`].
+    pub fn ecdsa_sign(&self, hash_value: &[u8]) -> Result<Vec<u8>, SymCryptError> {
+        // SymCrypt code AV's with SymCrypt Fatal Error.
+        // Panic'ing is not normal here in Rust so we are handling the error instead and returning SymCryptError::InvalidArgument
+        if self.get_ec_curve_usage() == EcKeyUsage::EcDh {
+            return Err(SymCryptError::InvalidArgument);
+        }
```
mlindgren, line 45: This should not cause a fatal error.

```
    // Make sure that the key may be used in ECDSA
    if ( ((pKey->fAlgorithmInfo & SYMCRYPT_FLAG_ECKEY_ECDSA) == 0)  )
    {
        scError = SYMCRYPT_INVALID_ARGUMENT;
        goto cleanup;
    }
``` #Resolved

File: rust-symcrypt/src/ecc/ecdsa.rs [hunk 13]
```diff
@@ -0,0 +1,231 @@
...
+        // SymCrypt code AV's with SymCrypt Fatal Error for trying to sign with Curve25519 as it is not supported.
+        // Panic'ing is not normal here in Rust so we are handling the error instead and returning SymCryptError::InvalidArgument
+        if self.get_curve_type() == CurveType::Curve25519 {
+            return Err(SymCryptError::InvalidArgument);
+        }
```
mlindgren, line 50: Nice find--this is arguably a bug in SymCrypt.

Curve25519 cannot be used with ECDSA, so it's expected that signing doesn't work. But I don't think it should fatal. As the comments say, SymCrypt is a "low-level" library so we tend to leave a lot of parameter validation to higher layers for performance reasons. But checking the curve type before signing (or perhaps at the time of key generation/import) would not be costly relative to the cost of the cryptographic operations.

Anyway, best to leave this here for now, but I'll file a bug on this. #Resolved

mlindgren (changes requested):
:clock1:

rust-symcrypt/src/lib.rs
nnmkhang: We have not update the version since we first released the crate. 

What I aim to do is do a major change with all of the updated code that we've added and make note of the breaking change. 

We would change from `0.2.0` to `0.3.0`, and version `0.2.0` would still work fine.

rust-symcrypt/src/ecc/ecdh.rs
nnmkhang: Changed

nnmkhang: Changed

rust-symcrypt/src/ecc/ecdsa.rs
nnmkhang: Thanks, I'll leave as is for now and we can file an issue and fix it if and when symcrypt fixes the bug

rust-symcrypt/src/ecc/mod.rs
nnmkhang: Export of the private key was not needed for the verify functions, I'll include it in this PR for completion

rust-symcrypt/src/ecc/mod.rs
nnmkhang: Changed to `public_key_length`

rust-symcrypt/src/ecc/ecdsa.rs
nnmkhang: Just tested this again with a debugger, in rust we panic here with `STATUS_STACK_BUFFER_OVERRUN`, will chat offline with you about this

rust-symcrypt/src/ecc/mod.rs
nnmkhang: at this point `ec_curve` is not null since we've done the check in `internal_new()`

To my knowledge, memory allocation error would only ever occur in this case if the entire system ran out of memory, which would cause Rust to crash anyways 


```
// Stack allocated since SymCryptEcCurveAllocate is called.
let curve_ptr = symcrypt_sys::SymCryptEcurveAllocate(to_symcrypt_curve(curve), 0);
if curve_ptr.is_null() {
  return Err(SymCryptError::MemoryAllocationFailure);
}
// Curve needs to be wrapped to properly free the curve in the case there is an error in future initialization in EcDsa or EcDh.
Ok(InnerEcCurve(curve_ptr))

```

rust-symcrypt/src/ecc/ecdh.rs
mlindgren: To clarify, this wasn't about the type, but about the language of the rest of the sentence. "The secret agreement between the private key and the public key" is not wrong, but it reads strangely to me, hence I'd just say that it "returns the agreed secret"

rust-symcrypt/src/ecc/mod.rs
mlindgren: That is true of pretty much all memory allocation failures - they're rare on modern systems with large amounts of RAM and virtual memory. But it's still good to handle them explicitly so that you know what the behavior is in the error case.

If the expected behavior is for the program to panic when it can't allocate these curves (which seems sensible given that they're "static" objects with respect to the lifetime of the program), I would recommend doing that explicitly in `InnerEcCurve::new`

rust-symcrypt/src/ecc/ecdsa.rs
mlindgren: Weird. I'll file another SymCrypt bug on that. You can leave it as-is for now.

rust-symcrypt/src/ecc/ecdh.rs
mlindgren: Still need to update the documentation comment here

rust-symcrypt/src/ecc/mod.rs
mlindgren: Cleared up via Teams chat - I misunderstood how this worked in Rust. The current way you're doing it is fine.

I do think removing these comments would be good; they just lead to confusion.

mlindgren (approved):
Couple remaining suggestions for updating doc comments. Otherwise, looks good!

PR #35, opened 2024-10-07
Bump to Version 0.3.0 

nnmkhang:
Bump to version 0.3.0 on cargo. This is a significantly big enough change to warrant a bump to v0.3.0 

What's been added:
- SymCrypt Version change from 103.1 -> 103.4.2
- Support for RSA ( pkcs1, oaep, pss ) 
- Support for EcDsa
- Support for Sha512 hashing
- Support for Sha3 hashing (Sha3_256, Sha3_384, Sha3_512)
- Support for Sha1, Md5 hashing ( disabled by default, only to be used for interop) 
- Support for Sha1, Md5 Hmac (disabled by default, only to be used for interop) 
- Added OS specific binding support for Windows and Linux on both AMD64 and ARM64

Version change:
Version 0.3.0's bindings are directly dependent on version `103.4.2`
The version of SymCrypt on the machine must be updated to one that is greater than 103.4.2 in order for the bindings to work.

Breaking Changes:
- Signature for EcDh has changed to account for a generic key type that is shared between EcDsa and EcDh
- `symcrypt_init()` has been set to private, and is called at every API entry point rather than passing the responsibility to the caller.
- Support for AL2 dynamic linking has been deprecated


rust-symcrypt/src/ecc/mod.rs
```diff
@@ -414,21 +416,16 @@ impl EcKey {
...
+    static ref CURVE_25519: Result<InnerEcCurve, SymCryptError> = internal_new(CurveType::Curve25519);
 }
 
 // SymCryptInit must be called before any EcDh operations are performed.
 fn internal_new(curve: CurveType) -> Result<InnerEcCurve, SymCryptError> {
     unsafe {
         // SAFETY: FFI calls
-        // Will only init once, subsequent calls to symcrypt_init() will be no-ops.
-        // Calling here incase user did not call symcrypt_init() earlier on.
-        symcrypt_init();
```
PhilHallin, line 430: >symcrypt_init(); [](http://example.com/codeflow?start=8&length=16)

Why remove the init? #Closed

nnmkhang: Because the entry point for creating the curve has to already call `symcrypt_init()`. 

Example: 

```
 pub fn generate_key_pair(
        curve_type: CurveType,
        ec_key_usage: EcKeyUsage,
    ) -> Result<Self, SymCryptError> {
        symcrypt_init();
        let ec_curve = InnerEcCurve::new(curve_type)?;
        unsafe {
               .... 
```

I don't think there are any scenarios where you'd be able to call this without `symcrypt_init()` already being called

mlindgren:
I'd recommend making the PR title more descriptive about the new features, e.g. "Version 0.3.0: add support for RSA, ECDSA, various hash algs"

rust-symcrypt/README.md
```diff
@@ -4,36 +4,19 @@ This crate provides friendly and idiomatic Rust wrappers over [SymCrypt](https:/
 
 This crate has a dependency on `symcrypt-sys`, which utilizes `bindgen` to create Rust/C FFI bindings.
 
-**Note:** As of version `0.2.0`, only `Windows AMD64`, and [`Azure Linux`](https://github.com/microsoft/azurelinux) are fully supported, with partial support for other Linux distros such as `Ubuntu`.
+**`symcrypt` version `0.3.0` is based off of `SymCrypt v103.5.0.1`.**
```
mlindgren: >v103.5.0.1 [](http://example.com/codeflow?start=55&length=10)

v103.5.1 #Resolved

rust-symcrypt/README.md
```diff
@@ -4,36 +4,19 @@ This crate provides friendly and idiomatic Rust wrappers over [SymCrypt](https:/
...
-
-For more information please see the `INSTALL.md` file on the [`rust-symcrypt`](https://github.com/microsoft/rust-symcrypt/tree/main/rust-symcrypt) page
-
-**Note:** This path may be different depending on your flavour of Linux. The goal is to place the `libsymcrypt.so*` files in a location where the your Linux distro can find the required libs at build/run time.
+| Operating Environment | Architecture      | Dynamic Linking |
+| --------------------- | ----------------- | ----------- |
+| Windows user mode     | AMD64, ARM64      | ✅          | 
+| Ubuntu (Tested via WSL)       | AMD64, ARM64      | ✅          | 
+| Azure Linux 3         | AMD64, ARM64      | ✅          |
+| Azure Linux 2         | AMD64, ARM64      | ❌          |
```
mlindgren: I'm not clear what this means. We don't have static linking support yet in this version, right? So Azl2 is just not supported at all, I'd think. #Resolved

rust-symcrypt/README.md
```diff
@@ -4,36 +4,19 @@ This crate provides friendly and idiomatic Rust wrappers over [SymCrypt](https:/
...
-
-Package manager support is already available if you are using `Azure Linux` via `tdnf install symcrypt`.
-
-For more information please see the `INSTALL.md` file on the [`rust-symcrypt`](https://github.com/microsoft/rust-symcrypt/tree/main/rust-symcrypt) page
-
-**Note:** This path may be different depending on your flavour of Linux. The goal is to place the `libsymcrypt.so*` files in a location where the your Linux distro can find the required libs at build/run time.
+| Operating Environment | Architecture      | Dynamic Linking |
+| --------------------- | ----------------- | ----------- |
+| Windows user mode     | AMD64, ARM64      | ✅          | 
+| Ubuntu (Tested via WSL)       | AMD64, ARM64      | ✅          |
```
mlindgren: >(Tested via WSL) [](http://example.com/codeflow?start=9&length=16)

Don't think it's necessary to mention that this was tested with WSL. WSL 2 is basically just a Linux VM so there should be no difference. #Resolved

rust-symcrypt/README.md
```diff
@@ -75,8 +59,47 @@ RSA:
 To enable either `Md5` or `Sha1` pass the `md5` or `sha1` flag into your `Cargo.toml`
 To enable all weak crypto, you can instead pass `weak-crypto` into your `Cargo.toml` instead.
 
+---
+
+
+## Quick Start Guide
+
+`symcrypt` requires the `SymCrypt` library to be present during build, and subsequently run time. The configuration of this dynamic link will differ from Windows/Linux.
```
mlindgren: >during build, and subsequently run time [](http://example.com/codeflow?start=57&length=39)

np: "at both build time and run time." #Resolved

mlindgren: >The configuration of this dynamic link will differ from Windows/Linux. [](http://example.com/codeflow?start=98&length=70)

np: I think this sentence is unnecessary  #Resolved

rust-symcrypt/README.md
```diff
@@ -75,8 +59,47 @@ RSA:
...
 
+---
+
+
+## Quick Start Guide
+
+`symcrypt` requires the `SymCrypt` library to be present during build, and subsequently run time. The configuration of this dynamic link will differ from Windows/Linux.
+
+### Windows:
+Download the latest `symcrypt.dll` and `symcrypt.lib` for you corresponding CPU architecture from the [SymCrypt Releases Page](https://github.com/microsoft/SymCrypt/releases) and place them somewhere accessible on your machine.
```
mlindgren: >you [](http://example.com/codeflow?start=58&length=3)

your #Resolved

rust-symcrypt/README.md
```diff
@@ -75,8 +59,47 @@ RSA:
...
+
+Set the required `SYMCRYPT_LIB_PATH` environment variable. You can do this by using the following command:
+
+`setx SYMCRYPT_LIB_PATH "<your-path-to-symcrypt-lib-folder>"`
+
+You will need to restart `terminal` / `cmd` after setting the environment variable.
+
+For more information please see the `INSTALL.md` file on the [`rust-symcrypt`](https://github.com/microsoft/rust-symcrypt/tree/main/rust-symcrypt) page
```
mlindgren: >page [](http://example.com/codeflow?start=147&length=4)

np: missing period #Resolved

rust-symcrypt/README.md
```diff
@@ -75,8 +59,47 @@ RSA:
...
+
+### Linux:
+
+#### Azure Linux 3:
+SymCrypt is pre-installed on Azure Linux 3 machines. Please ensure that you have the most up to date version of SymCrypt by updating via `tdnf`
+
+
+#### Other distros:
+
+Download the latest `libsymcrypt.so*` files for you corresponding CPU architecture from the [SymCrypt Releases Page](https://github.com/microsoft/SymCrypt/releases) and place them in your machines `$LD_LIBRARY_PATH`.
```
mlindgren: >you [](http://example.com/codeflow?start=48&length=3)

your #Resolved

rust-symcrypt/src/lib.rs
```diff
@@ -90,9 +10,8 @@ pub mod hash;
 pub mod hmac;
 pub mod rsa;
 
-/// `symcrypt_init()` must be called before any other function in the library. `symcrypt_init()` can be called multiple times,
-///  all subsequent calls will be no-ops
-pub fn symcrypt_init() {
+// symcrypt_init must be called before any other API can be called. All subsequent calls to symcrypt_init will be no-ops
+fn symcrypt_init() {
```
mlindgren, line 14: I was trying to find a better way to do this, because having to call `symcrypt_init` for every public function is very error prone. But it seems like there really isn't one built into the language, which seems a little crazy to me.

There is a crate that provides this functionality: [https://docs.rs/ctor/latest/ctor/](https://docs.rs/ctor/latest/ctor/)

There are a couple downsides to this though:
- The implementation is platform dependent (but seems to support all the platforms we need to, at least for now)
- Requires an additional crate dependency #Resolved

rust-symcrypt/src/ecc/ecdh.rs
```diff
@@ -39,6 +39,7 @@ use crate::errors::SymCryptError;
 use std::vec;
 use symcrypt_sys;
 
+#[allow(rustdoc::broken_intra_doc_links)]
```
mlindgren: Why? #Resolved

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -382,12 +397,12 @@ impl EcKey {
     }
 
     /// `get_curve_size()` returns a `u32` that represents the size of the curve associated with the [`EcKey`].
-    /// This call cannot fail.
-    pub fn get_curve_size(&self) -> u32 {
+    /// This call can fail with `SymCryptError::MemoryAllocationFailure` if there is not enough memory to allocate the curve.
+    pub fn get_curve_size(&self) -> Result<u32, SymCryptError> {
         unsafe {
-            let ec_curve = InnerEcCurve::new(self.get_curve_type());
+            let ec_curve = InnerEcCurve::new(self.get_curve_type())?;
```
mlindgren: Hm, you have `CurveType::get_size` now - why not just call that instead of allocating a new curve object just to get the size? #Resolved

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -440,14 +452,30 @@ fn internal_new(curve: CurveType) -> Result<InnerEcCurve, SymCryptError> {
...
-            CurveType::Curve25519 => &*CURVE_25519,
-        };
-        ec_curve
+    // new() returns an `InnerEcCurve` associated with the provided CurveType.
+    pub(crate) fn new(curve: CurveType) -> Result<&'static Self, SymCryptError> {
+        // Match the provided curve type and access the corresponding static `Result`.
+        match curve {
+            // Returning the lazy_static reference to the curve.
+            // Casting back our own MemoryAllocation error instead of getting the error from the static reference.
+            // This is done to avoid the need to clone the error.
```
mlindgren: Since this function also now returns a `Result<&'static Self, SymCryptError>`, can you not just directly return the static ref the same way you did before? The return type should be the same as the static ref type, no? #Resolved

rust-symcrypt/src/ecc/mod.rs
```diff
@@ -456,25 +484,30 @@ pub(crate) fn to_symcrypt_curve(curve: CurveType) -> symcrypt_sys::PCSYMCRYPT_EC
...
         CurveType::NistP256 => unsafe { symcrypt_sys::SymCryptEcurveParamsNistP256 }, // SAFETY: FFI calls
         CurveType::NistP384 => unsafe { symcrypt_sys::SymCryptEcurveParamsNistP384 }, // SAFETY: FFI calls
+        CurveType::NistP521 => unsafe { symcrypt_sys::SymCryptEcurveParamsNistP521 }, // SAFETY: FFI calls
         CurveType::Curve25519 => unsafe { symcrypt_sys::SymCryptEcurveParamsCurve25519 }, // SAFETY: FFI calls
     }
 }
 
 // curve_to_num_format() returns the correct number format needed for TLS interop since 25519 spec defines the use of Little Endian.
-pub(crate) fn curve_to_num_format(curve_type: CurveType) -> i32 {
+pub(crate) fn curve_to_num_format(curve_type: CurveType) -> symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT {
```
mlindgren: >_SYMCRYPT_NUMBER_FORMAT [](http://example.com/codeflow?start=74&length=23)

Not sure how typedefs work with bindgen, but can you not use the typedef'd `SYMCRYPT_NUMBER_FORMAT` (no underscore)? #Resolved

symcrypt-sys/src/lib.rs
```diff
@@ -1,9 +1,42 @@
...
+include!(concat!(
+    env!("CARGO_MANIFEST_DIR"),
+    "/bindings/linux_arm64_symcrypt_bindings.rs"
+));
+
+#[cfg(all(target_os = "windows", target_arch = "aarch64"))]
+include!(concat!(
+    env!("CARGO_MANIFEST_DIR"),
+    "/bindings/windows_arm64_symcrypt_bindings.rs"
+));
```
mlindgren, line 44: np: your comment above is grouped by OS but the macros are grouped by architecture. I'd prefer the former, but either is fine as long as it's consistent. Maybe separate the macro groups with the comments, e.g.

```
// Windows

...<macros>

// Linux

...<macros>
``` #Resolved

**Review by mlindgren — CHANGES_REQUESTED** [review]
:clock1:

File: rust-symcrypt/README.md (hunk 3 above)
Comment (nnmkhang) [reply rust-symcrypt/README.md]: Yeah thats true, I will remove the AL2 row

File: rust-symcrypt/src/lib.rs (hunk 9 above)
Comment (nnmkhang) [reply rust-symcrypt/src/lib.rs:14]: I tried using `ctor` but ran into issues with getting it to compile properly. We can chat offline about this

File: rust-symcrypt/src/ecc/ecdh.rs (hunk 10 above)
Comment (nnmkhang) [reply rust-symcrypt/src/ecc/ecdh.rs]: Auto generated for rust docs, removed

File: rust-symcrypt/src/ecc/mod.rs (hunk 11 above)
Comment (nnmkhang) [reply rust-symcrypt/src/ecc/mod.rs]: Changed to return the size from the `CurveType` impl. 

` pub fn get_curve_size(&self) -> u32 {  ... ` as this call can no longer fail

File: rust-symcrypt/src/ecc/mod.rs (hunk 12 above)
Comment (nnmkhang) [reply rust-symcrypt/src/ecc/mod.rs]: It has to do with the change in return types, and the fact that we are not using `unwrap()` anymore. We can't directly return back what we get from the `lazy_static` because the initialization of a `lazy_static` reference should not fail. Previously we were `unwrap()`ing to enforce that it wouldn't fail but now that we've added an Error variant `Result<InnerEcCurve, SymCryptError>` we have to handle with this match statement. I will change the code to be bit more concise though.

Comment (nnmkhang) [reply rust-symcrypt/src/ecc/mod.rs]: changed to `CurveType::NistP256 => NIST_P256.as_ref().map_err(|_| SymCryptError::MemoryAllocationFailure),` because its less verbose

File: rust-symcrypt/src/ecc/mod.rs (hunk 13 above)
Comment (nnmkhang) [reply rust-symcrypt/src/ecc/mod.rs]: Changed

File: symcrypt-sys/src/lib.rs (hunk 14 above)
Comment (nnmkhang) [reply symcrypt-sys/src/lib.rs:44]: Changed

File: rust-symcrypt/README.md (hunk 2 above)
Comment (nnmkhang) [reply rust-symcrypt/README.md]: Changed

File: rust-symcrypt/src/ecc/mod.rs (hunk 13 above)
Comment (mlindgren) [reply rust-symcrypt/src/ecc/mod.rs]: Don't need to fix this in this PR, but there are other places in the code where the _ prefix is still used. I think the only  reason that exists is because of historical quirks about how you used to have to use `typedef struct` in C. The non-underscore version is preferred, so if you have time it would be nice to clean the other occurrences up in a future PR, that would be nice.

File: rust-symcrypt/src/rsa/pkcs1.rs [hunk 15]
```diff
@@ -154,7 +154,10 @@ impl RsaKey {
                 0,
             ) {
                 symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(()),
-                err => Err(err.into()),
+                err => Err(match err.into() {
+                    SymCryptError::InvalidArgument => SymCryptError::SignatureVerificationFailure,
```
mlindgren: >SymCryptError::InvalidArgument => SymCryptError::SignatureVerificationFailure [](http://example.com/codeflow?start=20&length=77)

Why map this? We distinguish these error cases on purpose

rust-symcrypt/src/ecc/mod.rs
nnmkhang: Good callout, I'll put it as an action item for future PRs

rust-symcrypt/src/rsa/pkcs1.rs
PhilHallin: We are getting the InvalidArgument instead of SignatureVerificationFailure when passing in random signature bytes. From the caller's point of view we don't know why you are returning a different error?

samuel-lee-msft: It is useful to expose the distinction to the caller as _normally_ `SYMCRYPT_INVALID_ARGUMENT` would indicate a programmer error, and `SYMCRYPT_SIGNATURE_VERIFICATION_FAILURE` would indicate bad data.

In general `SYMCRYPT_INVALID_ARGUMENT` indicates something wrong about how the function was invoked (i.e. clearly bad arguments like flags set incorrectly, length of signature not matching the RSA key, etc.), and `SYMCRYPT_SIGNATURE_VERIFICATION_FAILURE` indicates that the input looks reasonable, but the signature does not verify.

I believe that you would currently hit `SYMCRYPT_INVALID_ARGUMENT` if a signature is going to fail regardless of the hash value you're trying to verify (i.e. if the signature value > RSA modulus then the signature is always invalid for that RSA key). This could typically happen when mixing up which RSA key a signature should be verified with; but it could also happen due to a corrupted signature.

Perhaps in your tests you could distinguish by signing a random hash, and then attempting to verify a perturbed hash with the same signature; this should always return `SYMCRYPT_SIGNATURE_VERIFICATION_FAILURE`.

rust-symcrypt/src/rsa/pss.rs
```diff
@@ -204,13 +207,13 @@ mod test {
             .unwrap();
 
         // tamper with signature
-        signature[0] = 0xFF;
+        signature[1] = 0xFF; // modifying second element, If we modify the first we will get InvalidArgument error instead.
```
samuel-lee-msft: I'd need to look into the details, but I think it's likely that some non-trivial amount of the time (probably less than 1/256, but triggerable occasionally) the signature will have 0xFF as the second byte already, and this won't be a modification.

When trying to test SignatureVerificationFailure, I would suggest two approaches:
1) Try to verify a signature signed by a distinct key of the same size
2) XOR a bit in the valid signature - i.e. this line becomes signature[1] ^= 0x1; - or something like that.

nnmkhang: Thanks for the advice, I will XOR instead

mlindgren (approved):
:shipit:

Issue #32, opened 2024-09-10
Intermittent test failures on pkcs1 and pss sign and verify 


nnmkhang:
When running `test_pkcs1_sign_and_verify_with_different_keys()`  the test is failing intermittently with `InvalidArgument` rather than `SignatureVerificationFailure`. This could be an issue with `generatge_key_pair()` creating a key with leading a `0`? 

also check `SymCryptRsaPcsa1ApplySignaturePAdding` 

possibly just force map `InvalidArugment` to `SignatureVerificationFailure`

PR #38, opened 2024-10-23
Intermittent test failures on pkcs1 and pss sign and verify  #32

MS-megliu:
When running test_pkcs1_sign_and_verify_with_different_keys() the test is failing intermittently with InvalidArgument rather than SignatureVerificationFailure. This could be an issue with generatge_key_pair() creating a key with leading a 0?

also check SymCryptRsaPcsa1ApplySignaturePAdding

possibly just force map InvalidArugment to SignatureVerificationFailure

Fix:
change the pss_verify and pkcs1_verify's return value, map invalidparameter to SignatureVerificationFailure. 

Test:
used cargo test pss and cargo test pkcs many times, worked okay.

### DIFF
--- a/rust-symcrypt/src/rsa/pkcs1.rs
+++ b/rust-symcrypt/src/rsa/pkcs1.rs
@@ -154,7 +154,10 @@ impl RsaKey {
                 0,
             ) {
                 symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(()),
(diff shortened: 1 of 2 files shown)


MS-megliu:
@microsoft-github-policy-service agree

MS-megliu:
@microsoft-github-policy-service agree company="Microsoft"

PhilHallin (approved):
:shipit:

PR #45, opened 2024-10-31
Add Aes Cbc encrypt/decrypt 

nnmkhang:
Added:
- Added AES CBC encrypt/decrypt 
- Added AES expanded key so that other encrypt/decrypt modes can be added in the future.
- Added bindgen build.rs to include new FFI for cbc 
- Updated bindings

Breaking changes:
- Moved `block_ciphers.rs` to `cipher/mod.rs`. Calling this would now need to be called from `use crate cipher::BlockCipherType` rather than `use crate::block_ciphers::*;`


rust-symcrypt/src/cipher/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    pub(crate) fn get_inner(&self) -> *const symcrypt_sys::SYMCRYPT_AES_EXPANDED_KEY {
+        &self.inner as *const _
+    }
+}
+
+/// `AesExpandedKey` is a struct that represents an expanded AES key. This struct holds no state and is used to encrypt and decrypt data.
+pub struct AesExpandedKey {
+    // Owned expanded key, this has no state, other calls will take reference to this key.
+    pub(crate) expanded_key: Arc<Pin<Box<AesInnerKey>>>,
```
nnmkhang: Should we give the user a peek at what the expanded key values are? or the key that they passed in? 

Something like: `get_key`? or `get_expanded_key`?  #Resolved

rust-symcrypt/src/cipher/cbc.rs
```diff
@@ -0,0 +1,319 @@
...
+//! # Examples
+//!
+//! ## AES-CBC Encryption in one go
+//!
+//! ```rust
+//! use symcrypt::cipher::AesExpandedKey;
+//! use hex::*;
+//!
+//! // Set up the key, chaining value, and plaintext
+//! let key = hex::decode("00000000000000000000000000000000").unwrap();
```
PhilHallin: >00000000000000000000000000000000 [](http://example.com/codeflow?start=27&length=32)

Something random #Closed

rust-symcrypt/src/cipher/cbc.rs
```diff
@@ -0,0 +1,319 @@
...
+//! let aes_cbc = AesExpandedKey::new(&key).unwrap();
+//! let mut chaining_value = hex::decode("00000000000000000000000000000000").unwrap();
+//! let mut plain_text = hex::decode("f34481ec3cc627bacd5dc3fb08f273e6").unwrap();
```
PhilHallin: >f34481ec3cc627bacd5dc3fb08f273e6 [](http://example.com/codeflow?start=38&length=32)

Something readable #Closed

rust-symcrypt/src/cipher/cbc.rs
```diff
@@ -0,0 +1,319 @@
...
+//! let mut chaining_value = hex::decode("db22065fb9302c4445151adc91310797").unwrap();
+//! let plaintext = hex::decode("4831f8d1a92cf167a444ccae8d90158dfc55c9a0742019e642116bbaa87aa205").unwrap();
+//! let expected_ciphertext = hex::decode("f03f86e1a6f1e23e70af3f3ab3b777fd43103f2e7a6fc245a3656799176a2611").unwrap();
+//!
+//! // Create the AES key
+//! let aes_cbc = AesExpandedKey::new(&key).unwrap();
+//!
+//! // Setup piecewise encrypt
+//! let mut piecewise_encrypted = vec![0u8; plaintext.len()];
+//! let mid = plaintext.len() / 2;
```
PhilHallin: Not a real pattern.

Better pattern is an array of fixed disk blocks to encrypt/decrypt. #Closed

rust-symcrypt/src/cipher/mod.rs
PhilHallin: I don't see a need at this time.

rust-symcrypt/src/cipher/cbc.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// If doing a piecewise encryption, you will need to keep track of the `chaining_value` and pass it in on the next block for encryption.
+    ///
+    /// `plain_text` is a `&[u8]` that represents the data to be encrypted.
+    ///
+    /// `cipher_text` is a `mut &[u8]` that will be filled with the encrypted data. The length of this buffer must be equal to the length of the `plain_text` buffer.
+    ///
+    /// This function will return a [`SymCryptError`] if the length of the `plain_text` and `cipher_text` buffers are not equal, or if they are not multiples of 16 bytes.
+    pub fn aes_cbc_encrypt(
+        &self,
+        chaining_value: &mut [u8],
```
PhilHallin: >chaining_value: &mut [u8], [](http://example.com/codeflow?start=8&length=26)

This must be the following length:
SYMCRYPT_AES_BLOCK_SIZE #Closed

rust-symcrypt/src/cipher/cbc.rs
```diff
@@ -0,0 +1,319 @@
...
+        plain_text: &[u8],
+        cipher_text: &mut [u8],
```
PhilHallin, line 84: What about in-place encryption. That would be a very common pattrn. #Closed

rust-symcrypt/src/cipher/cbc.rs
```diff
@@ -0,0 +1,319 @@
...
+//! let mut plain_text = hex::decode("f34481ec3cc627bacd5dc3fb08f273e6").unwrap();
+//! let mut cipher_text = vec![0u8; plain_text.len()];
+//!
+//! // Encrypt the plaintext
+//! aes_cbc.aes_cbc_encrypt(&mut chaining_value, &plain_text, &mut cipher_text).unwrap();
+//!
+//! assert_eq!(hex::encode(cipher_text), "0336763e966d92595a567cc9ce537f5e");
+//! ```
+//!
+//! ## AES-CBC Encryption piecewise
```
PhilHallin: >piecewise [](http://example.com/codeflow?start=26&length=9)

Its block encryption that can be called multiple times. #Closed

rust-symcrypt/src/cipher/cbc.rs
nnmkhang: Gcm docs do not show something readable, going to leave as is

rust-symcrypt/src/cipher/cbc.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// If doing encryption block by block, you will need to keep track of the `chaining_value` and pass it in on the next block for encryption. This value must be 16 bytes long.
+    ///
+    /// `plain_text` is a `&[u8]` that represents the data to be encrypted.
+    ///
+    /// `cipher_text` is a `mut &[u8]` that will be filled with the encrypted data. The length of this buffer must be equal to the length of the `plain_text` buffer.
+    ///
+    /// This function will return a [`SymCryptError`] if the length of the `plain_text` and `cipher_text` buffers are not equal, or if they are not multiples of 16 bytes.
+    pub fn aes_cbc_encrypt(
+        &self,
+        chaining_value: &mut [u8; 16],
```
PhilHallin: > chaining_value: &mut [u8; 16], [](http://example.com/codeflow?start=7&length=31)

Either a IV type or an IV length constant. #Closed

rust-symcrypt/README.md
```diff
@@ -43,13 +43,10 @@ HKDF:
...
 
-GCM:
-- Encryption ( in place )
-- Decryption ( in place )
-
-ChaCha:
-- Encryption ( in place )
-- Decryption ( in place )
+Encryption: 
+- Gcm Encrypt/Decrypt
```
samuel-lee-msft: NP: **AES**-GCM - there are GCM + other block ciphers defined (i.e. SM4-GCM)

rust-symcrypt/README.md
```diff
@@ -43,13 +43,10 @@ HKDF:
...
+- ChaCha Encrypt/Decrypt
```
samuel-lee-msft: NP: ChaCha20-Poly1305

rust-symcrypt/src/cipher/mod.rs
```diff
@@ -0,0 +1,192 @@
+//! Block Cipher functions related to creating expanded keys. For further information please see symcrypt.h for more info
```
samuel-lee-msft, line 1: why is this file called mod.rs?

rust-symcrypt/src/cipher/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+        return Err(SymCryptError::WrongDataSize);
+    }
+
+    // length of plain_text and cipher_text must be equal at this point.
+    if plain_text.len() % AES_BLOCK_SIZE as usize != 0 {
+        return Err(SymCryptError::WrongBlockSize);
+    }
+
+    Ok(())
+}
```
samuel-lee-msft: Hrmm, this is a bit of a strange function to have - not all block cipher modes have these restrictions on plaintext length vs. ciphertext length - and it's very odd to have this function use AES_BLOCK_SIZE for a function name that is not AES-specific.

Suggest moving this function into cbc.rs - and naming it validate_aes_cbc_inputs or something like that?

rust-symcrypt/src/cipher/cbc.rs
```diff
@@ -0,0 +1,330 @@
...
+        let mut cipher_text = vec![0u8; plain_text.len()];
+
+        aes_cbc
+            .aes_cbc_encrypt(&mut chaining_value, &plain_text, &mut cipher_text)
+            .unwrap();
+        assert_eq!(hex::encode(cipher_text), "0336763e966d92595a567cc9ce537f5e");
+    }
+
+    #[test]
+    fn test_aes_cbc_encrypt_single_and_piecewise() {
```
samuel-lee-msft, line 249: Would probably be good to also have a piecewise / block-wise decryption test

rust-symcrypt/README.md
mlindgren: Interestingly, the underlying SymCrypt functions expose this as a generic GCM block cipher mode, but it does look like only AES is (currently) supported. I agree that it's probably best to message it as just AES-GCM for now. But I wonder if it will be confusing that the Rust primitive refers to just GCM? In theory, it could be used with other ciphers in the future.

samuel-lee-msft (commented):
Overall looks reasonable to me :)

rust-symcrypt/README.md
```diff
@@ -43,13 +43,10 @@ HKDF:
...
+- Aes CBC Encrypt/Decrypt
```
mlindgren: >Aes CBC [](http://example.com/codeflow?start=2&length=7)

Nit: kind of a weird mix of capitalization in this file. Previous ones seem to match the struct names (Pascal case), but this one and the ECC/RSA algorithms below are more like how you'd write them in documentation (they're all acronyms).

If the latter, this should be AES-CBC. #Resolved

rust-symcrypt/README.md
samuel-lee-msft: Right - in SymCrypt API GCM was written as a mode over a block cipher - but in practice we only support AES-GCM (https://github.com/microsoft/SymCrypt/blob/907622c6658877014ea3fea336e2efa407864fdc/inc/symcrypt_internal.h#L1584-L1587)

nnmkhang: For the GCM code we have `pub fn new(key: &[u8], cipher: BlockCipherType) -> Result<Self, SymCryptError> {` where it takes in `BlockCipherType` 

I think that if AES is only supported and will only be supported in the near future, it would make sense to change GCM  to AES-GCM similar to how we have AES-CBC right now. If you all agree, I will mark it as a Todo and file it under our (growing) backlog

rust-symcrypt/src/cipher/mod.rs
nnmkhang: Its just a way to separate your code into folders so the `mod.rs` under `ciphers/mod.rs` will group all of the cipher related functions, similar to how we have `ecc/mod.rs` or `rsa/mod.rs`, it sort of acts like a mini library,

So when you call this code from an external crate or another part of the code base you would do `cipher::aes_expand_key` , or `rsa::generate_new_key_pair()`

rust-symcrypt/README.md
nnmkhang: Thanks for the catch, changing to AES-CBC

rust-symcrypt/src/cipher/cbc.rs
nnmkhang: Added

rust-symcrypt/src/cipher/mod.rs
nnmkhang: Moved to cbc.rs, renamed to `validate_aes_cbc_inputs`

rust-symcrypt/README.md
nnmkhang: changed

rust-symcrypt/README.md
mlindgren: That's fine with me. I think in the future it makes sense for symcrypt-sys to expose this the same way SymCrypt does, i.e. generic block cipher modes that can be used with any supported block cipher. I'm okay with making them AES-specific for now, but we should also be consistent between CBC and GCM.

rust-symcrypt/src/cipher/mod.rs
mlindgren: @samuel-lee-msft It's mod as in module, not modulus :)

rust-symcrypt/src/cipher/mod.rs
mlindgren: The format of the expanded key in SymCrypt is an implementation detail which callers should not rely on

rust-symcrypt/README.md
mlindgren: Just realized that the way this is implemented isn't consistent in SymCrypt either - ECB, CBC, and CTR have cipher-specific functions whereas GCM is generic. So that makes my suggestion a bit hard to do. Let's circle back on this later.

rust-symcrypt/src/cipher/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+// export ciphers
+pub mod cbc;
+
+/// 16
+pub const AES_BLOCK_SIZE: u32 = symcrypt_sys::SYMCRYPT_AES_BLOCK_SIZE;
+
+/// `BlockCipherType` is an enum that enumerates all possible block ciphers that are supported.
+/// Currently the only supported type is `AesBlock`.
+pub enum BlockCipherType {
+    AesBlock,
```
mlindgren, line 37: >AesBlock [](http://example.com/codeflow?start=4&length=8)

I would just call this `Aes`. AES is always a block cipher (although there are some block cipher modes that allow you to use them as stream ciphers).

rust-symcrypt/README.md
samuel-lee-msft: So for AES-KW(P) I am just exposing AES-KW(P) functionality directly.
This is similar to our existing AES-XTS functionality, where we only have an AES-specific flavor exposed.

ECB, CBC, and CtrMsb64 support both generic blockcipher and specific AES flavors.

GCM only supports the generic blockcipher, with the AES-GCM specific functionality being internal to SymCrypt.


I think I would argue for Rust-SymCrypt having only specific functions in the friendly layer, i.e. only support a subset of AES modes - do not support 3DES etc.
If we need to support some other block cipher in future we add new functions, not new enumeration values to a generic block cipher interface.

samuel-lee-msft: But yeah, only exposing specifically "AES-GCM" functions from Rust-SymCrypt can be a backlog item.
We should really make this change (expose AES-GCM specific functions) in the SymCrypt layer as well - the current situation in the SymCrypt API is confused.

mlindgren (approved):
lgtm modulo the one remaining suggestion to rename `AesBlock` to `Aes`

rust-symcrypt/src/cipher/mod.rs
nnmkhang: Im going to leave this one for now, changing it would be a breaking change since it would change the pub enum type which is used by GCM. WIll lump this in with the reformatting of AES-GCM

PR #46, opened 2024-11-01
Fixing pkcs1_decrypt bug

nnmkhang:
Fixing bug that breaks `pkcs1_decrypt` for Linux.

Breaking change:
- Updating function signature for `pkcs1_decrypt` from `u64` to `usize` to match the bindings for windows and Linux. 
- Updating to 0.4.0 since this is a breaking change. 
- This should only affect users that are on `0.3.0` and are using `pkcs1_decrypt` on Linux, otherwise no changes.

```
pub fn pkcs1_decrypt(
        &self,
        encrypted_message: &[u8],
        plaintext_buffer: &mut [u8],
        result_size: &mut usize,  <---- was previously result_size: &mut u64 
    ) -> SymCryptError {
```

### DIFF
--- a/rust-symcrypt/Cargo.lock
+++ b/rust-symcrypt/Cargo.lock
@@ -22,7 +22,7 @@ checksum = "9c198f91728a82281a64e1f4f9eeb25d82cb32a5de251c6bd1b5154d63a8e7bd"
 
 [[package]]
 name = "symcrypt"
-version = "0.3.0"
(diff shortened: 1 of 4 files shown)


PhilHallin (approved):
:shipit:

samuel-lee-msft (approved):
Comments offline about opportunity to tidy up a bit, but fine with this approach

PR #56, opened 2024-12-27
Replace symcrypt-bindgen crate with a script

Vaiz:
This PR introduces several significant changes to the project, focusing on replacing the symcrypt-bindgen crate with a script. The generated bindings remain unchanged, ensuring no modifications to the public API of symcrypt-sys.

### New scripts:

* Added `scripts/generate-bindings.ps1` to generate Rust bindings using bindgen for a specified target triple.
* Added `scripts/generate-all-bindings.ps1` to generate Rust bindings for all supported target triples, including installation instructions and prerequisites.
* Added `scripts/generate-bindings.sh` as a helper script to set the PATH to cargo binaries and call `generate-bindings.ps1`.

PR #57, opened 2025-01-14
Rework symcrypt-bindgen crate, add CI checks

### LINKED ISSUE #4: Add unit test validation to github actions
run a `cargo build` and `cargo test` as a github action before PR can be merged. 

This should be included as a precaution when checking in code to the main branch. Doing this github action will ensure that the code is able to build, all unit tests pass, and the docs test pass as well.

Github action spins up a new VM each time it's called, so it has no context on the needed `sycrypttestmodule.dll` and `symcrypttestmodule.lib` are. One possible solution is to check in a `.lib` and `.dll` file to the repo so that the github actions VM can have access to the binaries to run the tests.

extend for testing `Arm64`, `Amd64` for both windows and linux. Must test both dynamic linking and static linking. Must run both `cargo test` and `symcryptunittest.exe` for all cases.

### LINKED ISSUE #34: Update symcrypt-bindgen to cross compile
Use the crate `cross-rs` to spin up a vm that will create dynamic bindings for `arm64` `amd64` for windows and linux, 

Additionally, add script to append the `dylib` cfg to list of needed functions

### LINKED ISSUE #47: Update bindgen to be more robust
Bindgen is good not perfect, there are many improvements that need to be addressed:
- Bindings on windows are casting `size_t` to `u64` , should instead cast to `usize`, see https://docs.rs/bindgen/0.58.1/bindgen/struct.Builder.html#method.size_t_is_usize
- dylink for dynamic on windows should not be hand written
- Add script to get new bindings and apply needed patches ( dylink ) 
- tighten bindings to not expose not needed functions

Vaiz:
This PR replaces PR #56


scripts/generate-bindings.ps1
```diff
@@ -0,0 +1,165 @@
...
+$ErrorActionPreference = "Stop"
+$PSNativeCommandUseErrorActionPreference = $True
+
+# Init variables
+$SymCryptSysCrate = "$PSScriptRoot/../symcrypt-sys"
+$SymCryptSysCrate = $SymCryptSysCrate.Replace("\", "/")
+$wrapperHeader = "$SymCryptSysCrate/inc/wrapper.h"
+$targetName = $triple.Replace("-", "_")
+$bindingsFile = "$outDir/$targetName.rs"
+$rustVersion = "1.64.0"
```
nnmkhang: >ustVersion = "1.64.0" [](http://example.com/codeflow?start=2&length=21)

Should write a comment explaining the discussion we had around rust / bindgen versioning

nnmkhang:
> pub const SYMCRYPT_CODE_VERSION_API: u32 = 103;

I tested x86 bindings on Windows, WSL and AL3. everything is working as expected

---
Refers to: symcrypt-sys/src/bindings/aarch64_pc_windows_msvc.rs:3 in 4ab5d8f. [](commit_id = 4ab5d8f7d5701077a4f313ed35952139e104b84a, deletion_comment = False)

nnmkhang:
> # It is not intended for manual editing.

going forward should we check in our own cargo.lock? its re-compiled each time by the user right? 

---
Refers to: Cargo.lock:2 in 4ab5d8f. [](commit_id = 4ab5d8f7d5701077a4f313ed35952139e104b84a, deletion_comment = False)

scripts/generate-all-bindings.ps1
```diff
@@ -0,0 +1,53 @@
+# This script generates Rust bindings for the SymCrypt library for all supported target triples.
+# Prerequisites:
+# - The script must be run on Windows with WSL installed.
+# - LLVM and bindgen must be installed on both Windows and WSL.
+
+# Installation instructions:
```
nnmkhang, line 6: >Installation [](http://example.com/codeflow?start=2&length=12)

Have we considered cross-rs? 

This can spin up a docker container for each target triple and we would need very little setup aside from setting up the dependencies for cross.

We could write full rust and the script can be ran from linux/windows.

Additionally I think this would scale well in the case that symcrypt ever decides to add more bindings, ex: MacOS, MUSL etc.

What do you think?

scripts/generate-bindings.ps1
Vaiz: I changed the logic, now it automatically takes rust_version from symcrypt-sys crate metadata

scripts/generate-all-bindings.ps1
Vaiz: It's probably the best way forward, but I suggest to postpone that part as it can be done in parallel with other changes.

.github/workflows/builld.yml
```diff
@@ -0,0 +1,80 @@
...
+    - name: Debug build
+      run: cargo build --locked --verbose --target ${{ matrix.target }}
+    
+    - name: Release build
+      run: cargo build --release --locked --verbose --target ${{ matrix.target }}
+
+    - name: Run tests
+      if: matrix.run-tests
+      shell: pwsh
+      run: cargo test --locked --verbose --target ${{ matrix.target }}
```
nnmkhang: Can you please add `--all-features` for this test?

nnmkhang (changes requested):
Looks good! Thanks for your work on this

.github/workflows/builld.yml
Vaiz: done,
I also added missing tests for release configuration

nnmkhang (approved):
Thanks for your hard work!

PR #58, opened 2025-01-14
Fix dangling pointers in rsa::generate_key_pair and rsa::set_key_pair

sgalkin:
This commit fixes two issues
- current implementation of rsa::generate_key_pair may pass dangling pointer to SymCrypt if user provides exponent
- current implementation of rsa::set_key_pair passes dangling pointers to primes and primes length to SymCrypt


sgalkin:
@microsoft-github-policy-service agree company="Microsoft"

nnmkhang (approved):
Good catch, thank you for pointing this out.

PR #81, opened 2025-02-06
User/nnmkhang/static linking

nnmkhang:
co-authored by @Vaiz 

### ✅ Admin Checklist


symcrypt-sys/inc/static_LinuxDefault.c
```diff
@@ -0,0 +1,90 @@
...
+
+#include "wrapper.h"
+#include <sys/random.h>
+#include <errno.h>
+
+SYMCRYPT_ENVIRONMENT_POSIX_USERMODE;
+
+VOID
+SYMCRYPT_CALL
+SymCryptModuleInit(
```
nnmkhang: Exposing ` SymCryptModuleInit` in the static link that under the covers calls `SymCryptInit` so that we don't have to add `SymCryptInit` to the bindings. 

The reasoning for this is that it exposes the `SymCryptInit` API to the public and will symbol undefined errors if called from a dynamic instance of `symcrypt-sys`

symcrypt-sys/inc/static_LinuxDefault.c
```diff
@@ -0,0 +1,90 @@
...
+// RETURN VALUE         top
+//        On success, getrandom() returns the number of bytes that were
+//        copied to the buffer buf.  This may be less than the number of
+//        bytes requested via buflen if either GRND_RANDOM was specified in
+//        flags and insufficient entropy was present in the random source or
+//        the system call was interrupted by a signal.
+//        On error, -1 is returned, and errno is set to indicate the error.
+VOID 
+SYMCRYPT_CALL
+SymCryptRandom( PBYTE pbRandom, SIZE_T cbRandom) {
```
nnmkhang: In some cases `getrandom` cannot get enough entropy as per `sys/getrandom` documentation. We are looping until we fill the buffer to account for this edge case

symcrypt-sys/inc/symcrypt_internal_shared.inc
```diff
@@ -1,33 +1,33 @@
...
-#define SYMCRYPT_CODE_VERSION_API       103
-#define SYMCRYPT_CODE_VERSION_MINOR     8
-#define SYMCRYPT_CODE_VERSION_PATCH     0
-
-#if defined(DBG)
-#define SYMCRYPT_DEBUG 1
-#else
-#define SYMCRYPT_DEBUG 0
-#endif
+//
```
nnmkhang, line 1: This file would ideally be checked in by SymCrypt

symcrypt-sys/inc/static_LinuxDefault.c
```diff
@@ -0,0 +1,90 @@
+//
```
nnmkhang, line 1: This file would ideally be checked in and maintained by SymCrypt

symcrypt-sys/inc/static_WindowsDefault.c
```diff
@@ -0,0 +1,64 @@
+//
```
nnmkhang, line 1: This file would ideally be checked in and maintained by SymCrypt

symcrypt-sys/inc/buildInfo.h
```diff
@@ -1,8 +1,8 @@
-#include "symcrypt_internal_shared.inc"
-
-#define _SYMCRYPT_STRING_INT(a) #a
-#define _SYMCRYPT_STRING(a) _SYMCRYPT_STRING_INT(a)
-#define SYMCRYPT_BUILD_INFO_BRANCH    ""
-#define SYMCRYPT_BUILD_INFO_COMMIT    "2025-01-28T01:44:15+01:00_53be637"
-#define SYMCRYPT_BUILD_INFO_VERSION   _SYMCRYPT_STRING(SYMCRYPT_CODE_VERSION_API) "." _SYMCRYPT_STRING(SYMCRYPT_CODE_VERSION_MINOR) "." _SYMCRYPT_STRING(SYMCRYPT_CODE_VERSION_PATCH)
-#define SYMCRYPT_BUILD_INFO_TIMESTAMP "2025-01-31T15:53:40"
+#include "symcrypt_internal_shared.inc"
```
nnmkhang, line 1: This file would ideally be checked in and maintained by SymCrypt

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,368 @@
...
+            base_files.push("env_posixUserMode.c");
+            module_files.push("inc/static_LinuxDefault.c");
+        }
+        Triple::aarch64_unknown_linux_gnu => {
+            base_files.push("env_posixUserMode.c");
+            module_files.push("inc/static_LinuxDefault.c");
+        }
+    }
+
+    let asm_files = match options.triple() {
```
nnmkhang, line 332: The required `.asm` files would ideally be checked in and maintained by SymCrypt

symcrypt-sys/Cargo.toml
```diff
@@ -6,13 +6,20 @@ license = "MIT OR Apache-2.0"
...
 homepage = "https://github.com/microsoft/SymCrypt"
 repository = "https://github.com/microsoft/rust-symcrypt"
 readme = "README.md"
-exclude = ["symcrypt/*", "inc/*"]
+exclude = ["symcrypt/*", "inc/*"] # FIXME: update this before merging to main branch
+#links = "symcrypt" # FIXME: uncomment this before merging to main branch
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
```
nnmkhang, line 18: This new `dynamic` feature will be enabled when dynamically linking. We should also put a compliance flag out there that enforces `dynamic` to be enabled for release builds

rust-symcrypt/Cargo.toml
```diff
@@ -24,6 +24,7 @@ default = []
 md5 = []
 sha1 = []
 pkcs1-encrypt-decrypt = []
+dynamic = ["symcrypt-sys/dynamic"]
```
nnmkhang, line 27: This is the feature that most developers would be interacting with. They would set the `dynamic` feature in the `rust-symcrypt` layer, and the `dynamic` flag would propagate down to the `symcrypt-sys` layer.

rust-symcrypt/src/lib.rs
```diff
@@ -17,6 +17,7 @@ fn symcrypt_init() {
     static INIT: Once = Once::new();
     unsafe {
         // SAFETY: FFI calls, blocking from being run again.
+        // Under the covers, the default static lib implementation exposes SymCryptModuleInit, but calls SymCryptInit for static linking
```
nnmkhang: What are you thoughts on this approach? it provided a safer `symcrypt-sys` API on our end since we don't expose `SymCryptInit` to `symcrypt-sys`. We could use `[cfg]` to handle this, but it that adds some complexity to the management of features and IMO will cause issues maintaining long-term.

nnmkhang: See more on the comments i've put on `static_WindowsDefault.c`

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,422 @@
...
+                cc.flag("-mpclmul");
+                cc.flag("-Wno-incompatible-pointer-types");
+                /*
+                cc.flag("-mpclmul")
+                    .flag("-mssse3")
+                    .flag("-mxsave")
+                    .flag("-maes")
+                    .flag("-msha")
+                    .flag("-mrdrnd")
+                    .flag("-mrdseed");
```
mamckee, line 144: Why comment these out? The flags should match the platform-specific flags SymCrypt is normally build with: [https://github.com/microsoft/SymCrypt/blob/53be637dab201a4c9d95e1cf58040c85d71cf3c2/cmake-configs/SymCrypt-Platforms.cmake#L75](https://github.com/microsoft/SymCrypt/blob/53be637dab201a4c9d95e1cf58040c85d71cf3c2/cmake-configs/SymCrypt-Platforms.cmake#L75) #Resolved

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,422 @@
+use super::triple::Triple;
```
nnmkhang, line 1: Majority of the code change is within this file

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,422 @@
...
+                cc.flag("-mpclmul")
+                    .flag("-mssse3")
+                    .flag("-mxsave")
+                    .flag("-maes")
+                    .flag("-msha")
+                    .flag("-mrdrnd")
+                    .flag("-mrdseed");
+                */
+            }
+            Triple::aarch64_unknown_linux_gnu => {
```
mamckee, line 146: Same as above. This is missing platform-specific flags:
`-march=armv8-a+simd+crypto` and `-flax-vector-conversions`

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,422 @@
...
+    triple: Triple,
+    symcrypt_use_asm: bool,
+    //symcrypt_fips_build: bool, // TODO: Determine if we should expose FIPS build option?
+}
+
+impl SymCryptOptions {
+    fn new() -> Self {
+        Self {
+            triple: Triple::get_target_triple(),
+            symcrypt_use_asm: false, // FIXME: Turn this to true when we get ASM checked in
```
nnmkhang, line 57: As mentioned in a meeting before we don't want to explicitly say that a certain build is not FIPS compliant, but in this scenario I think it could make sense alerting customers that static linking is strictly not FIPS.

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,422 @@
...
+                /*
+                cc.flag("-mpclmul")
+                    .flag("-mssse3")
+                    .flag("-mxsave")
+                    .flag("-maes")
+                    .flag("-msha")
+                    .flag("-mrdrnd")
+                    .flag("-mrdseed");
+                */
+            }
```
nnmkhang, line 145: Wondering if anyone has context on the flags i may or may not be missing?

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,422 @@
...
+            .include("symcrypt/lib")
+            .warnings(false); // Ignore noisy warnings from SymCrypt
+
+        if !self.symcrypt_use_asm {
+            cc.define("SYMCRYPT_IGNORE_PLATFORM", None);
+        }
+
+        // Set specific flags for each target
+        match self.triple {
+            Triple::x86_64_pc_windows_msvc => {
```
mamckee, line 128: This should define `_AMD64_` #Resolved

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,422 @@
...
+            Triple::aarch64_unknown_linux_gnu => {
+                cc.include("symcrypt/modules/linux/common"); // TODO: remove?
+                cc.flag("-Wno-incompatible-pointer-types"); // Ignore noisy SymCrypt errors
+            }
+        }
+
+        cc
+    }
+}
+
```
nnmkhang, line 156: TODO in a future PR, have a check that alerts the rust-symcrypt dev that the CMAKE file has been updated.

symcrypt-sys/inc/static_LinuxDefault.c
```diff
@@ -0,0 +1,93 @@
...
```
nnmkhang, line 1: Open to name suggestions, not strong on `static_XXXDefault.c`

symcrypt-sys/build/static_link.rs
nnmkhang: Added

symcrypt-sys/src/bindings/x86_64_unknown_linux_gnu.rs
```diff
@@ -4788,9 +4793,14 @@ extern "C" {
 extern "C" {
     pub fn SymCryptHkdfSelfTest();
 }
+#[cfg(feature = "dynamic")]
```
nnmkhang, line 4796: `SymCryptRandom` is not compiled unless `dynamic` feature is set. This applies to all target bindings

symcrypt-sys/src/bindings/x86_64_unknown_linux_gnu.rs
```diff
@@ -4198,6 +4198,11 @@ extern "C" {
 extern "C" {
     pub fn SymCryptStoreMsbFirstUint64(src: UINT64, pbDst: PBYTE, cbDst: SIZE_T) -> SYMCRYPT_ERROR;
 }
+#[cfg(not(feature = "dynamic"))]
```
nnmkhang, line 4201: `SymCryptModuleInit()` is only exposed when `dynamic` feature is present. This applies for all triples

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,433 @@
...
+        if !self.symcrypt_use_asm {
+            cc.define("SYMCRYPT_IGNORE_PLATFORM", None); // TODO: Fix when we get ASM
+        }
+
+        // Set specific flags for each target
+        match self.triple {
+            Triple::x86_64_pc_windows_msvc => {
+                cc.define("_AMD64_", None);
+                // cc.asm_flag("/DSYMCRYPT_MASM"); // MASM assembly TODO: enable for ASM
+                cc.flag("/MP"); // Multi-threaded compilation
```
mamckee: These flags should be set for Windows aarch64 too

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,433 @@
...
+                cc.flag("-mpclmul");
+                cc.flag("-Wno-incompatible-pointer-types"); // Ignore noisy SymCrypt errors
+                                                            // From SymCrypt-Platforms.cmake
+                cc.flag("-mssse3");
+                cc.flag("-mxsave");
+                cc.flag("-maes");
+                cc.flag("-msha");
+                cc.flag("-mrdrnd");
+                cc.flag("-mrdseed");
+                cc.flag("-mbmi2");
```
mamckee: >cc.flag("-mbmi2"); [](http://example.com/codeflow?start=16&length=18)

This is only set for sha256-ymm.c and sha512-ymm.c in the SymCrypt build. You set this below so I don't think this is needed here.

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,433 @@
...
+                cc.flag("/WX"); // Treat warnings as errors
+                cc.flag("/guard:cf"); // Control Flow Guard
+                                      // cc.flag("/dynamicbase"); // Enable ASLR
+                cc.flag("/EHsc"); // Exception handling
+            }
+            Triple::aarch64_pc_windows_msvc => {
+                cc.define("_ARM64_", None);
+            }
+            Triple::x86_64_unknown_linux_gnu => {
+                cc.flag("-mpclmul");
```
mamckee: >cc.flag("-mpclmul"); [](http://example.com/codeflow?start=16&length=20)

nit: This should go with the other instruction set flags

symcrypt-sys/build/static_link.rs
mamckee: Still missing these arguments that are shared with AMD64 build: [https://github.com/microsoft/SymCrypt/blob/53be637dab201a4c9d95e1cf58040c85d71cf3c2/cmake-configs/SymCrypt-Platforms.cmake#L119](https://github.com/microsoft/SymCrypt/blob/53be637dab201a4c9d95e1cf58040c85d71cf3c2/cmake-configs/SymCrypt-Platforms.cmake#L119)

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,433 @@
...
+        Triple::x86_64_pc_windows_msvc => vec![
+            "aesasm.asm",
+            "fdef_asm.asm",
+            "fdef_mulx.asm",
+            "fdef369_asm.asm",
+            "sha256xmm_asm.asm",
+            "sha256ymm_asm.asm",
+            "sha512ymm_asm.asm",
+            "sha512ymm_avx512vl_asm.asm",
+            "wipe.asm",
```
mamckee: I think these should be the *-gas.asm files

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,433 @@
...
+        ],
+        Triple::aarch64_pc_windows_msvc => vec!["fdef_asm.asm", "fdef369_asm.asm", "wipe.asm"],
```
mamckee: >["fdef_asm.asm", "fdef369_asm.asm", "wipe.asm"], [](http://example.com/codeflow?start=47&length=48)

I think these are supposed to be *-gas.asm too

symcrypt-sys/build/static_link.rs
```diff
@@ -0,0 +1,433 @@
...
+                                                            // From SymCrypt-Platforms.cmake
+                cc.flag("-mssse3");
+                cc.flag("-mxsave");
+                cc.flag("-maes");
+                cc.flag("-msha");
+                cc.flag("-mrdrnd");
+                cc.flag("-mrdseed");
+                cc.flag("-mbmi2");
+                cc.flag("-fstack-protector-strong");
+                cc.flag("-Wstack-protector");
```
mamckee: Missing "--param ssp-buffer-size=4". Also why not enable stack protection for all targets?

nnmkhang: I think stack protection only applies to linux:

https://github.com/microsoft/SymCrypt/blob/53be637dab201a4c9d95e1cf58040c85d71cf3c2/CMakeLists.txt#L79C1-L83C8

```
option(
    SYMCRYPT_STACK_PROTECTION
    "When enabled, SymCrypt will enable various stack protection mechanisms to protect against buffer overruns and the like. Only
    applicable to non-Windows systems. Defaults to ON, but may need to be disabled for specialized targets such as embedded systems."
    ON)
```

symcrypt-sys/build/static_link.rs
nnmkhang: I was copying the files from `lib/CMakeList.txt`, Is there somewhere else that we should be compiling it from?

https://github.com/microsoft/SymCrypt/blob/53be637dab201a4c9d95e1cf58040c85d71cf3c2/lib/CMakeLists.txt#L230C1-L266C54

```
if(WIN32 AND SYMCRYPT_USE_ASM)
    if(SYMCRYPT_TARGET_ARCH MATCHES "AMD64")
              ...

        list(APPEND SOURCES_COMMON
            amd64/aesasm-masm.asm
            amd64/fdef_asm-masm.asm
            amd64/fdef369_asm-masm.asm
            amd64/fdef_mulx-masm.asm
            amd64/wipe-masm.asm
            amd64/sha256xmm_asm-masm.asm
            amd64/sha256ymm_asm-masm.asm
            amd64/sha512ymm_asm-masm.asm
            amd64/sha512ymm_avx512vl_asm-masm.asm)
        set_source_files_properties(
            amd64/aesasm-masm.asm
            amd64/fdef_asm-masm.asm
            amd64/fdef369_asm-masm.asm
            amd64/fdef_mulx-masm.asm
            amd64/wipe-masm.asm
            amd64/sha256xmm_asm-masm.asm
            amd64/sha256ymm_asm-masm.asm
            amd64/sha512ymm_asm-masm.asm
            amd64/sha512ymm_avx512vl_asm-masm.asm
            PROPERTIES LANGUAGE ASM_MASM)
            ...
```

nnmkhang: Either way this will probably change significantly once we get the actual `.asm` files

symcrypt-sys/build/static_link.rs
nnmkhang: same as other [comment ](https://github.com/microsoft/rust-symcrypt/pull/81/files/8d5c9a2e9b80daa3d6953ac70a5ae11d40a7a77b#r1956656326)

symcrypt-sys/build/static_link.rs
nnmkhang: Spoke offline, adding  gas

rust-symcrypt/src/lib.rs
```diff
@@ -15,23 +15,50 @@ pub mod rsa;
...
+
+        #[cfg(not(feature = "dynamic"))]
+        INIT.call_once(|| {
+            symcrypt_sys::SymCryptInit();
+        });
+        // call symcrypt Init, only expose this for static linking. Add a regex to add a cfg feature to the bindings for this.
     }
 }
 
 /// Takes in a a buffer called `buff` and fills it with random bytes. This function cannot fail.
```
samuel-lee-msft: >This function cannot fail. [](http://example.com/codeflow?start=70&length=26)

I think it would be more accurate to say that it is never expected to fail, but failure (due to OS dependencies) will crash the application. There is no recoverable failure mode.

samuel-lee-msft: >a a [](http://example.com/codeflow?start=13&length=3)

NP: a

symcrypt-sys/build/static_link.rs
samuel-lee-msft: Currently for the SymCrypt build we have:
Windows build is only supported with MSVC, which uses the MASM-style (Microsoft Macro Assembler) assembly for AMD64 / x86, and ARMASM-style (Microsoft ARM assembler) for Arm64.

Linux / other builds is only supported with GNU CC-like tooling, which uses GAS (GNU assembly syntax)

PhilHallin (approved):
:shipit:

PR #96, opened 2026-04-24
User/nnmkhang/poc bcrypt

nnmkhang:
### ✅ Admin Checklist


rust-symcrypt/src/backend/mod.rs
```diff
@@ -0,0 +1,5 @@
+#[cfg(windows)]
+pub mod bcrypt;
```
mlindgren, line 2: I feel like rather than having the implementation chosen automatically based on the platform, it should be a feature flag. Of course, BCrypt is only available on Windows, but we could give Windows callers the choice of calling through BCrypt or linking to SymCrypt.

rust-symcrypt/src/backend/symcrypt/gcm.rs
```diff
@@ -0,0 +1,135 @@
+use crate::errors::SymCryptError;
```
mlindgren, line 1: I'm assuming code in this directory was just moved and is (mostly) unchanged

rust-symcrypt/src/backend/bcrypt/gcm.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+// BCrypt can fail here (NTSTATUS); SymCrypt has no equivalent 
+fn open_algorithm(alg_id: *const u16, flags: u32) -> AlgHandle {
+    let mut handle: BCRYPT_ALG_HANDLE = std::ptr::null_mut();
+    let _status = unsafe {
+        BCryptOpenAlgorithmProvider(&mut handle, alg_id, std::ptr::null(), flags)
+    };
+    AlgHandle(handle)
```
mlindgren: Seems like we could just always use pseudo-handles instead of explicitly opening a handle

rust-symcrypt/src/backend/bcrypt/gcm.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+static AES_GCM_ALG: LazyLock<AlgHandle> = LazyLock::new(|| {
+    let handle = open_algorithm(BCRYPT_AES_ALGORITHM, 0);
+    // BCrypt can fail here (NTSTATUS); SymCrypt has no equivalent, block cipher is a static pointer.
+    let gcm_mode_bytes = 32u32;
+    let _status = unsafe {
```
mlindgren: >let _status = unsafe { [](http://example.com/codeflow?start=4&length=22)

Addendum to the above comment about psuedo-handles: there is a `BCRYPT_AES_GCM_ALG_HANDLE`; I believe if you use this you won't have to set the chaining mode.

rust-symcrypt/src/backend/bcrypt/gcm.rs
```diff
@@ -0,0 +1,171 @@
...
+// BCrypt can fail here (NTSTATUS); SymCrypt has no equivalent 
+fn open_algorithm(alg_id: *const u16, flags: u32) -> AlgHandle {
+    let mut handle: BCRYPT_ALG_HANDLE = std::ptr::null_mut();
+    let _status = unsafe {
+        BCryptOpenAlgorithmProvider(&mut handle, alg_id, std::ptr::null(), flags)
+    };
+    AlgHandle(handle)
+}
+
+static AES_GCM_ALG: LazyLock<AlgHandle> = LazyLock::new(|| {
```
samuel-lee-msft: Recommend using the `BCRYPT_AES_GCM_ALG_HANDLE` pseudohandle (https://learn.microsoft.com/en-us/windows/win32/seccng/cng-algorithm-pseudo-handles), which effectively does this internally in BCrypt

rust-symcrypt/src/backend/bcrypt/hash.rs
```diff
@@ -0,0 +1,565 @@
...
+});
+
+static SHA512_ALG: LazyLock<AlgHandle> = LazyLock::new(|| {
+    open_algorithm(BCRYPT_SHA512_ALGORITHM, BCRYPT_HASH_REUSABLE_FLAG)
+});
+
+// BCrypt SHA-3 algorithm IDs — not in windows-sys, requires Windows 11 24H2+
+const BCRYPT_SHA3_256_ALGORITHM: &[u16] = &[0x53, 0x48, 0x41, 0x33, 0x2D, 0x32, 0x35, 0x36, 0x00];
+const BCRYPT_SHA3_384_ALGORITHM: &[u16] = &[0x53, 0x48, 0x41, 0x33, 0x2D, 0x33, 0x38, 0x34, 0x00];
+const BCRYPT_SHA3_512_ALGORITHM: &[u16] = &[0x53, 0x48, 0x41, 0x33, 0x2D, 0x35, 0x31, 0x32, 0x00];
```
mlindgren, line 48: Would be nice if these could be represented as human readable strings somehow

rust-symcrypt/src/backend/bcrypt/gcm.rs
samuel-lee-msft: Ah duplicated https://github.com/microsoft/rust-symcrypt/pull/96#discussion_r3251642932 - will let Mitch do in-depth review.

rust-symcrypt/src/backend/symcrypt/gcm.rs
nnmkhang: Yeah just moved to a different spot, identical code

rust-symcrypt/src/backend/bcrypt/hash.rs
nnmkhang: did some playing around i have some options:

A: This would be the simplest to do
```
const BCRYPT_SHA3_256_ALGORITHM: &[u16] = &[
     b'S' as u16, b'H' as u16, b'A' as u16, b'3' as u16,
     b'-' as u16, b'2' as u16, b'5' as u16, b'6' as u16,
     0,
 ];
``` 

B: We have an addtional dep. now. Would need to pull in windows and windows-rs, not a big deal I think but another dep. to manage. 
```
use windows::core::w;
 const BCRYPT_SHA3_256_ALGORITHM: PCWSTR = w!("SHA3-256");
```

C: Keep as is, just add a comment saying that its encoded

rust-symcrypt/src/backend/mod.rs
nnmkhang: I initially thought about that as well, but there are some issues with adding a feature flag, mostly due to the (maybe not that likely) problem of multiple crates depending on `rust-symcrypt` at one time within the same cargo chain. Since cargo features are additive things could get messy very fast. Additionally, I don't think many users actually care how they get their crypto (`symcrypt` or `bcrypt`) they just want it to work and be in compliance. I don't know how many users would actually go out of their way to manually configure `symcrypt` linking, and adding that functionality can footgun us. 

Here are some of the options I have been thinking about: 

A: User opts into `symcrypt`, `bcrypt` is the default 
If a downstream user opts into `symcrypt`, then the whole chain is forced to also opt into `symcrypt` since the features are additive. This would introduce a run time dep.  if there was multiple apps depending on `rust-symcrypt`. This would be pretty bad IMO. 

B: User opts into `bcrypt`, `symcrypt` is the default 
If we have `symcrypt` with the dll as the default, and then have a user opt into `bcrypt`. This will always work since `bcrypt` is on every windows machine (ignoring PQC and SHA3 for a moment). In the worst case of merging, there would be additional env. variables for the dynamic library but it would not break any builds. Downside is batteries are not included. I think that most people would just turn on `bcrypt` anyways since they wouldn't want to deal with `symcrypt.dll`, which at that point, why even have them opt in at all? 

C: We select `bcrypt` by default on windows, no more dll, batteries are included. Still have flexibility to add other modes in the future, but this solves the use case for majority of users. 

D: static linking default, user opts into `bcrypt` for FIPS. 
By default we statically link on `windows` and `linux`, this makes it batteries included and we do not require any DLL stuff. We can let users on `windows` into `bcrypt` (FIPS mode) which is always on `windows` machines so does not introduce any additional runtime dep. Any other caller in the chain also flips onto FIPS mode and calls `bcrypt` as well. There is no breaking change here and no additional env. variables set. On `linux`, similar story but we document that you need to pull down the `dll` for FIPS, which at this point we have established as a reasonable requirement for non AL3 distos. 

In the future if we decide to ship `symcrypt.dll` in box, the switch under the covers for FIPS mode from `Bcrypt->SymCrypt.dll` does not need to be known for the caller. Or if we include FIPS for static linking, we can also do that without any breaking changes. This approach can be a bandaid if we decide to eventually certify static linking, which ties in with the `SymCrust` work. 

E: dynamically ship `symcrypt.dll` problem is now more or less gone from `rust-symcrypt`, and becomes a distribution problem within ES. This approach is extremely simple from the rust side.


Sorry, I was initially just going to lay out option A and B, but kind of got carried away.

rust-symcrypt/src/backend/bcrypt/hash.rs
mlindgren: What's the difference between `windows-sys` and `windows`/`windows-rs`? Can we call BCrypt via the latter two crates?

rust-symcrypt/src/backend/mod.rs
mlindgren: From a crate usability perspective, I like option D. I think the main issue with it is whether we can enforce internal users always using the FIPS feature.

PR #97, opened 2026-04-24
Update build.rs to support OpenVMM use cases

smalis-msft:
### Description of Changes:
This PR adds two features to build.rs. The first is supporting per-target environment variables for configuration. This aids in cross-compilation, and is the approach used by the openssl crate. The second is a new environment variable, SYMCRYPT_STATIC, that emits a static linking directive instead of a dynamic one. It does not add any support for producing a suitable static library. Finally this PR updates the documentation to include the new support.

### Breaking Changes if any:
I believe there should be none.

### DIFF
--- a/rust-symcrypt/INSTALL.md
+++ b/rust-symcrypt/INSTALL.md
@@ -1,10 +1,12 @@
 # Detailed Build and Install
 
-This page provides more detailed installation instructions for dynamic linking on Windows and Linux.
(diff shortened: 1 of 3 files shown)


nnmkhang (approved):
LGTM, I think it would be nice to have an example for others for static library setup, but I think that we can hold off until we have openVMM working, and provide a link as an example once its complete.

PR #99, opened 2026-05-08
add xts-aes and aes-kw(p) algos

### LINKED ISSUE #53: Support AES XTS
Hello,

The bindings here are missing support for AES-XTS. [This is supported in SymCrypt](https://github.com/microsoft/SymCrypt/blob/main/lib/xtsaes.c).

At first glance (admittedly very shallow), such support doesn't look terribly difficult to implement.

nnmkhang:
### Description of Changes:

- adding Aes_kw_xts
- adding AES-KW(P) 

### ✅ Admin Checklist
- [x] Review the PR description and ensure all necessary details are included.
- [x] Update/add unit tests for changed code.
- [x] Update/add documentation for new or changed APIs.
- [x] Run `cargo test --all-features` on `Windows` and `WSL`.

### DIFF
--- a/rust-symcrypt/README.md
+++ b/rust-symcrypt/README.md
@@ -59,11 +59,6 @@ Key Wrap (NIST SP 800-38F):
 - AES-KW (RFC 3394) — wraps 8-byte-aligned plaintext, output is plaintext + 8 bytes
 - AES-KWP (RFC 5649) — wraps any non-zero-length plaintext
 
-**Note**: AES-KW(P) has no general-purpose equivalent in the Windows BCrypt / CNG API
-(BCrypt only exposes the RFC 3565 key-blob format via `BCryptExportKey`, and KWP not at
-all). A future BCrypt-backed variant of this crate is expected to return a `NotSupported`
-runtime error for AES-KW(P) on Windows until that gap is addressed upstream.
-
 ECC:
 - ECDH Secret Agreement ( NistP256, NistP384, NistP521, Curve25519)
 - ECDSA Sign / Verify ( NistP256, NistP384, NistP521 )
--- a/rust-symcrypt/README.md
+++ b/rust-symcrypt/README.md
@@ -53,6 +53,16 @@ Encryption:
 - AES-GCM Encrypt/Decrypt
 - ChaCha20-Poly1305 Encrypt/Decrypt
 - AES-CBC Encrypt/Decrypt
+- XTS-AES-256 Encrypt/Decrypt
+
+Key Wrap (NIST SP 800-38F):
+- AES-KW (RFC 3394) — wraps 8-byte-aligned plaintext, output is plaintext + 8 bytes
+- AES-KWP (RFC 5649) — wraps any non-zero-length plaintext
+
+**Note**: AES-KW(P) has no general-purpose equivalent in the Windows BCrypt / CNG API
+(BCrypt only exposes the RFC 3565 key-blob format via `BCryptExportKey`, and KWP not at
+all). A future BCrypt-backed variant of this crate is expected to return a `NotSupported`
+runtime error for AES-KW(P) on Windows until that gap is addressed upstream.
 
 ECC:
 - ECDH Secret Agreement ( NistP256, NistP384, NistP521, Curve25519)
--- a/rust-symcrypt/README.md
+++ b/rust-symcrypt/README.md
@@ -53,7 +53,7 @@ Encryption:
 - AES-GCM Encrypt/Decrypt
 - ChaCha20-Poly1305 Encrypt/Decrypt
 - AES-CBC Encrypt/Decrypt
-- XTS-AES-256 Encrypt/Decrypt
+- XTS-AES Encrypt/Decrypt (128, 192, 256-bit keys)
 
 Key Wrap (NIST SP 800-38F):
 - AES-KW (RFC 3394) — Wraps plaintext that is a multiple of 8 bytes. Output length is the length of the plaintext + 8 bytes.
--- a/rust-symcrypt/README.md
+++ b/rust-symcrypt/README.md
@@ -56,7 +56,7 @@ Encryption:
 - XTS-AES-256 Encrypt/Decrypt
 
 Key Wrap (NIST SP 800-38F):
-- AES-KW (RFC 3394) — wraps 8-byte-aligned plaintext, output is plaintext + 8 bytes
+- AES-KW (RFC 3394) — Wraps plaintext that is a multiple of 8 bytes. Output length is the length of the plaintext + 8 bytes.
 - AES-KWP (RFC 5649) — wraps any non-zero-length plaintext
 
 ECC:
--- a/rust-symcrypt/README.md
+++ b/rust-symcrypt/README.md
@@ -43,6 +43,12 @@ HKDF:
 - HmacSha384
 - HmacSha512
 
+SP800-108 Counter Mode KDF:
+- HmacSha1
+- HmacSha256
+- HmacSha384
+- HmacSha512
+
 Encryption:
 - AES-GCM Encrypt/Decrypt
 - ChaCha20-Poly1305 Encrypt/Decrypt
--- a/rust-symcrypt/src/cipher/aes_kw.rs
+++ b/rust-symcrypt/src/cipher/aes_kw.rs
@@ -38,14 +38,12 @@
 //! assert_eq!(recovered, plaintext);
 //! ```
 
-use crate::cipher::AesInnerKey;
+use crate::cipher::{expand_aes_key, AesInnerKey};
 use crate::errors::SymCryptError;
-use crate::symcrypt_init;
 use std::pin::Pin;
 use symcrypt_sys;
 
 const KW_SEMIBLOCK: usize = 8;
-const KW_MIN_PLAINTEXT_LEN: usize = 16;
 
 /// [`AesKwKey`] wraps an expanded AES key for use with AES-KW.
 pub struct AesKwKey {
@@ -57,28 +55,11 @@ pub struct AesKwpKey {
     expanded_key: Pin<Box<AesInnerKey>>,
 }
 
-fn expand_aes_key(key: &[u8]) -> Result<Pin<Box<AesInnerKey>>, SymCryptError> {
-    let mut expanded_key = AesInnerKey::new();
-    unsafe {
-        // SAFETY: FFI call. Returns WrongKeySize for non-AES key lengths.
-        match symcrypt_sys::SymCryptAesExpandKey(
-            expanded_key.as_mut().get_inner_mut(),
-            key.as_ptr(),
-            key.len() as symcrypt_sys::SIZE_T,
-        ) {
-            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(expanded_key),
-            err => Err(err.into()),
-        }
-    }
-}
-
 impl AesKwKey {
     /// `new()` returns an [`AesKwKey`] or a [`SymCryptError`] if the key is the wrong size.
     /// Accepted key sizes are 16, 24, or 32 bytes (AES-128/192/256).
     pub fn new(key: &[u8]) -> Result<Self, SymCryptError> {
-        symcrypt_init();
-        let expanded_key = expand_aes_key(key)?;
-        Ok(AesKwKey { expanded_key })
+        Ok(AesKwKey { expanded_key: expand_aes_key(key)? })
     }
 
     /// `encrypt()` wraps `plaintext` using AES-KW and returns the ciphertext.
@@ -86,10 +67,6 @@ impl AesKwKey {
     /// `plaintext.len()` must be a multiple of 8 and at least 16. The returned `Vec` has length
     /// `plaintext.len() + 8`.
     pub fn encrypt(&self, plaintext: &[u8]) -> Result<Vec<u8>, SymCryptError> {
-        symcrypt_init();
-        if plaintext.len() < KW_MIN_PLAINTEXT_LEN || plaintext.len() % KW_SEMIBLOCK != 0 {
-            return Err(SymCryptError::WrongDataSize);
-        }
         let mut ciphertext = vec![0u8; plaintext.len() + KW_SEMIBLOCK];
         let mut written: symcrypt_sys::SIZE_T = 0;
         unsafe {
@@ -116,12 +93,6 @@ impl AesKwKey {
     /// `ciphertext.len()` must be a multiple of 8 and at least 24. Returns
     /// [`SymCryptError::AuthenticationFailure`] if the integrity check fails.
     pub fn decrypt(&self, ciphertext: &[u8]) -> Result<Vec<u8>, SymCryptError> {
-        symcrypt_init();
-        if ciphertext.len() < KW_MIN_PLAINTEXT_LEN + KW_SEMIBLOCK
-            || ciphertext.len() % KW_SEMIBLOCK != 0
-        {
-            return Err(SymCryptError::WrongDataSize);
-        }
         let mut plaintext = vec![0u8; ciphertext.len() - KW_SEMIBLOCK];
         let mut written: symcrypt_sys::SIZE_T = 0;
         unsafe {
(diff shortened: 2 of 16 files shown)


smalis-msft (approved):

rust-symcrypt/README.md
```diff
@@ -53,6 +53,11 @@ Encryption:
 - AES-GCM Encrypt/Decrypt
 - ChaCha20-Poly1305 Encrypt/Decrypt
 - AES-CBC Encrypt/Decrypt
+- XTS-AES-256 Encrypt/Decrypt
+
+Key Wrap (NIST SP 800-38F):
+- AES-KW (RFC 3394) — wraps 8-byte-aligned plaintext, output is plaintext + 8 bytes
```
mlindgren: >wraps 8-byte-aligned plaintext, output is plaintext + 8 bytes [](http://example.com/codeflow?start=22&length=61)

Wraps plaintext that is a multiple of 8 bytes. Output length is the length of the plaintext + 8 bytes. #Resolved

rust-symcrypt/src/cipher/aes_kw.rs
```diff
@@ -0,0 +1,363 @@
...
+            key.len() as symcrypt_sys::SIZE_T,
+        ) {
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(expanded_key),
+            err => Err(err.into()),
+        }
+    }
+}
+
+impl AesKwKey {
+    /// `new()` returns an [`AesKwKey`] or a [`SymCryptError`] if the key is the wrong size.
```
mlindgren, line 86: >[`AesKwKey`] or a [`SymCryptError`] [](http://example.com/codeflow?start=27&length=35)

np: why are these in square brackets? IIRC cargo-doc automatically links symbols #Resolved

rust-symcrypt/src/cipher/aes_kw.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    /// `encrypt()` wraps `plaintext` using AES-KW and returns the ciphertext.
+    ///
+    /// `plaintext.len()` must be a multiple of 8 and at least 16. The returned `Vec` has length
+    /// `plaintext.len() + 8`.
+    pub fn encrypt(&self, plaintext: &[u8]) -> Result<Vec<u8>, SymCryptError> {
+        symcrypt_init();
+        if plaintext.len() < KW_MIN_PLAINTEXT_LEN || plaintext.len() % KW_SEMIBLOCK != 0 {
+            return Err(SymCryptError::WrongDataSize);
+        }
```
mlindgren: The plaintext length is validated in `SymCryptAesKwEncrypt`, but it returns `SYMCRYPT_INVALID_ARGUMENT`.

I think it's okay to add additional checks in places where SymCrypt doesn't have them, but where they already exist, we should leave the validation up to the underlying function. #Resolved

rust-symcrypt/src/cipher/aes_kw.rs
```diff
@@ -0,0 +1,363 @@
...
+        let expanded_key = expand_aes_key(key)?;
+        Ok(AesKwKey { expanded_key })
+    }
+
+    /// `encrypt()` wraps `plaintext` using AES-KW and returns the ciphertext.
+    ///
+    /// `plaintext.len()` must be a multiple of 8 and at least 16. The returned `Vec` has length
+    /// `plaintext.len() + 8`.
+    pub fn encrypt(&self, plaintext: &[u8]) -> Result<Vec<u8>, SymCryptError> {
+        symcrypt_init();
```
mlindgren: >symcrypt_init(); [](http://example.com/codeflow?start=8&length=16)

This seems probably unnecessary - you cannot call this function without having an `AesKwKey` object first, and callers outside of this module can't instantiate `AesKwKey` without calling `new` first.

Likewise for decrypt #Resolved

rust-symcrypt/src/cipher/aes_kw.rs
```diff
@@ -0,0 +1,363 @@
...
+pub struct AesKwKey {
+    expanded_key: Pin<Box<AesInnerKey>>,
+}
+
+/// [`AesKwpKey`] wraps an expanded AES key for use with AES-KWP.
+pub struct AesKwpKey {
+    expanded_key: Pin<Box<AesInnerKey>>,
+}
+
+fn expand_aes_key(key: &[u8]) -> Result<Pin<Box<AesInnerKey>>, SymCryptError> {
```
mlindgren: This is duplicated from `AesExpandKey` #Resolved

rust-symcrypt/src/cipher/aes_kw.rs
```diff
@@ -0,0 +1,363 @@
...
+                    plaintext.truncate(written as usize);
+                    Ok(plaintext)
+                }
+                err => Err(err.into()),
+            }
+        }
+    }
+}
+
+impl AesKwpKey {
```
mlindgren, line 155: >AesKwpKey [](http://example.com/codeflow?start=5&length=9)

Same comments here as for `AesKwpKey` #Resolved

rust-symcrypt/src/cipher/xts.rs
```diff
@@ -0,0 +1,477 @@
...
+use std::mem;
+use std::pin::Pin;
+use std::ptr;
+use symcrypt_sys;
+
+const XTS_AES_256_KEY_LEN: usize = 64;
+const XTS_DATA_UNIT_MIN: u64 = 16;
+const XTS_DATA_UNIT_MAX: u64 = 1 << 24;
+
+/// [`XtsAes256Key`] holds the expanded XTS-AES-256 key from SymCrypt.
```
mlindgren: >[`XtsAes256Key`] [](http://example.com/codeflow?start=4&length=16)

Same comment here about square brackets #Resolved

rust-symcrypt/src/cipher/xts.rs
```diff
@@ -0,0 +1,477 @@
...
+            _pinned: PhantomPinned,
+        })
+    }
+
+    /// Provides a mutable pointer to the inner SymCrypt state.
+    ///
+    /// This is primarily meant to be used while making calls to the underlying SymCrypt APIs.
+    /// The pointer returned is pinned and cannot be moved
+    /// This function returns pointer to pinned data, which means callers must not use the pointer to move the data out of its location.
+    fn get_inner_mut(self: Pin<&mut Self>) -> *mut symcrypt_sys::SYMCRYPT_XTS_AES_EXPANDED_KEY {
```
mlindgren, line 96: >fn get_inner_mut [](http://example.com/codeflow?start=4&length=16)

I wonder if this should be an unsafe function? And maybe it should be called `get_inner_mut_ptr` #Resolved

rust-symcrypt/src/cipher/xts.rs
```diff
@@ -0,0 +1,477 @@
+//! XTS-AES-256 (NIST SP 800-38E) functions. For further documentation please refer to symcrypt.h
```
mlindgren: >XTS-AES-256 [](http://example.com/codeflow?start=4&length=11)

SymCrypt supports 128 and 192 bit keys for XTS-AES as well. Is there a specific reason we don't want to support those here? #Resolved

mlindgren (dismissed):
:clock1:

rust-symcrypt/src/cipher/aes_kw.rs
nnmkhang: For cargo doc to link you need to use the `[]`, just plain ``` only does the formatting

rust-symcrypt/README.md
nnmkhang: Will update

rust-symcrypt/src/cipher/xts.rs
nnmkhang: Added,

rust-symcrypt/src/cipher/xts.rs
nnmkhang: The other functions in the repo that have similar are unsafe. If we want to make it unsafe, I propose a follow-up PR where we do it to all of them

rust-symcrypt/src/cipher/aes_kw.rs
```diff
@@ -0,0 +1,331 @@
...
+        })
+    }
+
+    /// `encrypt()` wraps `plaintext` using AES-KW and returns the ciphertext.
+    ///
+    /// `plaintext.len()` must be a multiple of 8 and at least 16. The returned `Vec` has length
+    /// `plaintext.len() + 8`.
+    pub fn encrypt(&self, plaintext: &[u8]) -> Result<Vec<u8>, SymCryptError> {
+        let mut ciphertext = vec![0u8; plaintext.len() + KW_SEMIBLOCK];
+        let mut written: symcrypt_sys::SIZE_T = 0;
```
mlindgren, line 102: **[security/low] Integer overflow in output-buffer sizing.**

```rust
let mut ciphertext = vec![0u8; plaintext.len() + KW_SEMIBLOCK];
```

On a 32-bit target, `plaintext.len() + 8` can wrap on adversarially-sized inputs and produce an undersized output buffer that is then passed to `SymCryptAesKwEncrypt` with `ciphertext.len()` — SymCrypt would reject it (no UB), but `next_multiple_of` in the KWP variant (lines ~131-132) will panic on overflow in release as well. Prefer `checked_add` and return `SymCryptError::InvalidArgument` on overflow so both modes fail uniformly without panicking. #Resolved

rust-symcrypt/src/cipher/aes_kw.rs
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    /// `decrypt()` unwraps `ciphertext` using AES-KW and returns the recovered plaintext.
+    ///
+    /// `ciphertext.len()` must be a multiple of 8 and at least 24. Returns
+    /// [`SymCryptError::AuthenticationFailure`] if the integrity check fails.
+    pub fn decrypt(&self, ciphertext: &[u8]) -> Result<Vec<u8>, SymCryptError> {
+        // SymCrypt will handle invalid data sizes
+        let mut plaintext = vec![0u8; ciphertext.len()];
```
mlindgren: **[security/low] Decrypt buffer over-allocated by 8 bytes.**

`vec![0u8; ciphertext.len()]` — the recovered plaintext is always `ciphertext.len() - 8` for AES-KW (and at most that for KWP). Functionally safe (zero-initialized, truncated after the call), but trims attack surface and avoids a fixed leak window into the tail if SymCrypt ever wrote past `written` in a future revision. Prefer `ciphertext.len().saturating_sub(KW_SEMIBLOCK)`. Same comment applies to the KWP `decrypt`. #Resolved

rust-symcrypt/src/cipher/aes_kw.rs
```diff
@@ -0,0 +1,331 @@
...
+                }
+                err => Err(err.into()),
+            }
+        }
+    }
+
+    /// `decrypt()` unwraps `ciphertext` using AES-KW and returns the recovered plaintext.
+    ///
+    /// `ciphertext.len()` must be a multiple of 8 and at least 24. Returns
+    /// [`SymCryptError::AuthenticationFailure`] if the integrity check fails.
```
mlindgren, line 125: **[security/medium] Decrypted plaintext is not zeroized on drop.**

`decrypt` returns a plain `Vec<u8>`; on drop, residual plaintext sits in the allocator until reuse. The expanded key state is correctly wiped via `SymCryptWipe`, but recovered key material — the primary use of AES-KW(P)! — is not. Consider returning a wrapper that wipes on drop, or at minimum document that callers should wrap with `zeroize`. This is broader than this PR, but AES-KW(P) is uniquely a key-protection primitive, so it matters more here.

rust-symcrypt/src/cipher/aes_kw.rs
```diff
@@ -0,0 +1,331 @@
...
+                }
+                err => Err(err.into()),
+            }
+        }
+    }
+}
+
+impl AesKwpKey {
+    /// `new()` returns an [`AesKwpKey`] or a [`SymCryptError`] if the key is the wrong size.
+    /// Accepted key sizes are 16, 24, or 32 bytes (AES-128/192/256).
```
mlindgren, line 157: **[security/low — doc bug] KWP `encrypt` doc-comment is incorrect and uses C ternary.**

```text
plaintext.len() + 16 - (plaintext.len() % 8) - ((plaintext.len() % 8) == 0 ? 8 : 0)
```

Not valid Rust, and it doesn't match the implementation, which is `plaintext.len().next_multiple_of(8) + 8`. Doc/spec mismatch in a crypto API is a security-relevant smell — please fix the formula. #Resolved

rust-symcrypt/src/cipher/xts.rs
```diff
@@ -0,0 +1,688 @@
...
+    pub fn decrypt_in_place_with_128b_tweak(
+        &self,
+        data_unit_size: u64,
+        tweak: &[u8; 16],
+        buffer: &mut [u8],
+    ) -> Result<(), SymCryptError> {
+        validate_xts_args(data_unit_size, buffer.len())?;
+        unsafe {
+            // SAFETY: FFI call.
+            symcrypt_sys::SymCryptXtsAesDecryptWith128bTweak(
```
mlindgren, line 378: **[security/medium] `data_unit_size` validation accepts non–block-aligned values.**

Per SP 800-38E, an XTS data unit is a multiple of the AES block size (16). The current validator accepts e.g. `data_unit_size = 17`. Either tighten the check (`data_unit_size % 16 == 0`) or remove the partial range check entirely and fully delegate to SymCrypt — the current state is a partial check that may mislead callers into thinking they're validated when they aren't. #Resolved

rust-symcrypt/src/cipher/xts.rs
```diff
@@ -0,0 +1,688 @@
...
+    /// 128-bit tweak.
+    ///
+    /// `data_unit_size` is the size in bytes of each data unit, typically 512.
+    /// It must be at least 16 and at most 2^24. `buffer.len()` must be a multiple of `data_unit_size`.
+    /// `tweak` is a `&[u8; 16]` containing the 128-bit tweak for the first data unit; it is
+    /// incremented as a 128-bit integer for each subsequent data unit.
+    pub fn decrypt_in_place_with_128b_tweak(
+        &self,
+        data_unit_size: u64,
+        tweak: &[u8; 16],
```
mlindgren, line 372: **[security/low — error reporting] Validation ordering produces a misleading error.**

A caller passing `data_unit_size = 0` with an empty buffer gets `WrongDataSize` (from the `data_len == 0` check) instead of `InvalidArgument` (which describes the actual problem — the data unit size). Re-order so the `data_unit_size` range check runs before the `data_len == 0` check; that way each error variant points at the truly bad argument. #Resolved

mlindgren (commented):
LLM reviewer found some additional issues:

rust-symcrypt/src/cipher/xts.rs
nnmkhang: Added

rust-symcrypt/src/cipher/xts.rs
nnmkhang: Added to the validation check

rust-symcrypt/src/cipher/aes_kw.rs
nnmkhang: Changed to plaintex.len().next_multiple

rust-symcrypt/src/cipher/aes_kw.rs
nnmkhang: Added zeroize here, SymCrypt wipe does similar so we could technically extend from that code, but zeroized is pretty popular so theres no need to re-invent the wheel here IMO. 

Wondering human thought here, adding zeroize, makes the ergonomics a bit more annoying to use from the callers perspective, Is leaking info here a big deal?

If the leak is a big deal, should we also add to other APIs that create private keys? IE: ECDH?

smalis-msft: I would prefer not forcing this on consumers at this point in time. IMO this should be done everywhere or nowhere, and treated as a separate breaking change if we want to make it everywhere. As an alternative, and would probably be cleaner, you could take an &mut [u8] as an input, and the caller can choose whether it should be inside of a Zeroize or not. That would probably be my ideal preference. Then you can just return a Result<(), SymCryptError>. This is a pretty common pattern, so it wouldn't be considered unusual.

mlindgren: Sorry, that was an AI comment; I just skimmed it and made sure it was valid, but I should have added some extra context.

I think @smalis-msft makes a good point. This needs to be consistent between APIs. With the SymCrypt C APIs, it's always the caller's responsibility to wipe returned secret data when they are done with it (because there is no other option in C). In Rust, I think it would be nice to try to prevent the caller from shooting themselves in the foot, _if_ it can be done without making the APIs obnoxious to use. But it's not something I have done a lot of investigation into. This is also something we're still figuring out the best way to do in SymCRust, in addition to ensuring that we wipe all stack-allocated secret buffers.

For now, let's just document that it's the caller's responsibility to wipe secrets on drop. We can revisit in the future.

smalis-msft: My answer to that is that it isn't necessarily "obnoxious" in Rust, but currently all solutions would force the caller into accepting whatever solution this crate decides on, as there's no standard library solution. I would vote for leaving it to the caller unless and until some sort of ecosystem-wide consensus is reached, and perhaps not even then. Leaving something like this to the caller, and just taking as input an &mut [u8], is a much more common pattern. Or perhaps an &mut [u8;size] when the size of an output is known ahead of time, or an &mut Vec\<u8\> when it isn't. All valid choices.

mlindgren: Thanks for the info. IMO no need to change the API surface right now; just document that callers must wipe the `Vec` after use, and file a task to address this across the crate.

mlindgren (approved):
lgtm modulo the one remaining discussion about Zeroize

rust-symcrypt/src/cipher/aes_kw.rs
nnmkhang: Sounds good, I will remove the zeroize and track adding it across the crate as an action item

smalis-msft (approved):
shipit

PR #106, opened 2026-07-09
User/khangnguyen/mscrypto traits

nnmkhang:
### ✅ Admin Checklist

### DIFF
--- a/mscrypto-bcrypt/src/hash.rs
+++ b/mscrypto-bcrypt/src/hash.rs
@@ -1,0 +1,338 @@
+//! Hashing for the BCrypt provider
+
+use std::ptr;
+
+use mscrypto::algorithm::BaseHashAlgorithm;
+#[cfg(feature = "sha3")]
+use mscrypto::algorithm::Sha3Algorithm;
+use mscrypto::hash::provider::BuildDigest;
+use mscrypto::hash::{Digest, Hash, HashOps};
+
+use windows_sys::Win32::Foundation::NTSTATUS;
+use windows_sys::Win32::Security::Cryptography::*;
+
+use crate::BcryptProvider;
+
+const SHA256_LEN: usize = 32;
+const SHA384_LEN: usize = 48;
+const SHA512_LEN: usize = 64;
+
+// windows-sys exposes the SHA-2 algorithm pseudo-handles but not the SHA-3 ones, so
+// define them the same way it defines its own (a BCRYPT_ALG_HANDLE built from the value
+// in the bcrypt pseudo-handle table). SHA-3 pseudo-handles exist on Windows 11 24H2 and later.
+#[cfg(feature = "sha3")]
+const BCRYPT_SHA3_256_ALG_HANDLE: BCRYPT_ALG_HANDLE = 0x0000_03B1_u32 as _;
+#[cfg(feature = "sha3")]
+const BCRYPT_SHA3_384_ALG_HANDLE: BCRYPT_ALG_HANDLE = 0x0000_03C1_u32 as _;
+#[cfg(feature = "sha3")]
+const BCRYPT_SHA3_512_ALG_HANDLE: BCRYPT_ALG_HANDLE = 0x0000_03D1_u32 as _;
+
+// Mirrors the SDK's BCRYPT_SUCCESS / NT_SUCCESS macro: a nonnegative NTSTATUS is
+// success; only negative values are failures.
+fn bcrypt_success(status: NTSTATUS) -> bool {
+    status >= 0
+}
+
+/// Panics on a non-success `NTSTATUS`. Used on the infallible SHA-2 path (and once
+/// SHA-3 is running), where a failure is catastrophic and cannot be reported.
+fn expect_success(call: &str, status: NTSTATUS) {
+    if !bcrypt_success(status) {
+        panic!(
+            "mscrypto-bcrypt: {} failed: NTSTATUS 0x{:08X}",
+            call, status as u32
+        );
+    }
+}
+
+/// Owns a hash object handle, destroyed on drop.
+struct HashHandle(BCRYPT_HASH_HANDLE);
+
+unsafe impl Send for HashHandle {}
+
+impl Drop for HashHandle {
+    fn drop(&mut self) {
+        // SAFETY: handle came from BCryptCreateHash and is destroyed once.
+        unsafe { BCryptDestroyHash(self.0) };
+    }
+}
+
+/// Streaming hasher shared by the SHA-2 and SHA-3 surfaces. Both run off process-global
+/// algorithm pseudo-handles, so a hasher owns only its hash object.
+pub struct BcryptHasher {
+    hash: HashHandle,
+    output_len: usize,
+}
+
+impl BcryptHasher {
+    /// Builds a hasher from an algorithm pseudo-handle.
+    fn new(alg: BCRYPT_ALG_HANDLE, output_len: usize) -> Self {
+        BcryptHasher {
+            hash: create_hash_handle(alg),
+            output_len,
+        }
+    }
+}
+
+/// Creates a hash object from an algorithm pseudo-handle. Passing NULL/0 for the object
+/// buffer allocates it and free it via BCryptDestroyHash. Panics on failure.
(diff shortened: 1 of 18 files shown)


nnmkhang:
Hey Sam, large PR so some notes that might help:

This is the order I think you should look:
1. Contract for the traits from mscrypto. Specifically at `proivder.rs`, this is what `bcrypt` and `symcrypt` have to do in order to be used as a provider. The distinction here is that this is an owned object thats pulled in as a dependency rather than a feature. This solves most of the cargo clobbering issues that we were worried about with the feature design. 
2. Sha3 feature in the `mscrypto/cargo.toml` this feature enables/disables `sha3`, we will also do the same for each PQC algorithm as well.
3. `mscrypto-symcrypt`, this will be the most familiar for both for both of us, I'm just fulfilling the contract from the traits defined in step 1. 
4. `mscrypto-bcrypt`, this is the tricker part because of the bcrypt APIs, we check if the sha3 handle is available via `sha3_available`. This is used in the `lib.rs` for `mscrypto-bcrypt` to probe during build as well. 

I will leave other notes inline at the call sites. 

I also made a small sample program that shows how the provider would be chosen etc. 

RE: the discussion on `Sha3`. wondering your thoughts, It might just be easier to have `SHAX` return a `?`, its a bit ugly, but makes the maintenance slightly easier for us. I am still on the fence about this tbh.

mscrypto-bcrypt/src/hash.rs
```diff
@@ -0,0 +1,338 @@
...
+use crate::BcryptProvider;
+
+const SHA256_LEN: usize = 32;
+const SHA384_LEN: usize = 48;
+const SHA512_LEN: usize = 64;
+
+// windows-sys exposes the SHA-2 algorithm pseudo-handles but not the SHA-3 ones, so
+// define them the same way it defines its own (a BCRYPT_ALG_HANDLE built from the value
+// in the bcrypt pseudo-handle table). SHA-3 pseudo-handles exist on Windows 11 24H2 and later.
+#[cfg(feature = "sha3")]
```
nnmkhang, line 23: will switch these once they land in windows.rs

mscrypto-bcrypt/src/hash.rs
```diff
@@ -0,0 +1,338 @@
...
+    let mut handle: BCRYPT_HASH_HANDLE = ptr::null_mut();
+    // SAFETY: alg is a valid algorithm pseudo-handle; the hash object is bcrypt owned.
+    let status =
+        unsafe { BCryptCreateHash(alg, &mut handle, ptr::null_mut(), 0, ptr::null_mut(), 0, 0) };
+    expect_success("BCryptCreateHash", status);
+    HashHandle(handle)
+}
+
+impl HashOps for BcryptHasher {
+    fn update(&mut self, data: &[u8]) {
```
nnmkhang, line 88: Is this over-engineered? I dont think anyone is going to ever hash a big file like this in one shot

mscrypto-bcrypt/src/hash.rs
```diff
@@ -0,0 +1,338 @@
...
+    fn hash(&self, algorithm: BaseHashAlgorithm) -> BcryptHasher {
+        let (alg, len) = match algorithm {
+            BaseHashAlgorithm::Sha256 => (BCRYPT_SHA256_ALG_HANDLE, SHA256_LEN),
+            BaseHashAlgorithm::Sha384 => (BCRYPT_SHA384_ALG_HANDLE, SHA384_LEN),
+            BaseHashAlgorithm::Sha512 => (BCRYPT_SHA512_ALG_HANDLE, SHA512_LEN),
+        };
+        BcryptHasher::new(alg, len)
+    }
+
+    fn digest(&self, algorithm: BaseHashAlgorithm, data: &[u8]) -> Digest {
```
nnmkhang, line 125: Could one shot here with the bcrypt API, but see the other note about large hashes

mscrypto-bcrypt/src/hash.rs
```diff
@@ -0,0 +1,338 @@
...
+
+// Mirrors the SDK's BCRYPT_SUCCESS / NT_SUCCESS macro: a nonnegative NTSTATUS is
+// success; only negative values are failures.
+fn bcrypt_success(status: NTSTATUS) -> bool {
+    status >= 0
+}
+
+/// Panics on a non-success `NTSTATUS`. Used on the infallible SHA-2 path (and once
+/// SHA-3 is running), where a failure is catastrophic and cannot be reported.
+fn expect_success(call: &str, status: NTSTATUS) {
```
nnmkhang, line 38: Thoughts on this panic idea. I assume that any error (aside from unavailable for sha3) would be catastrophic enough to panic anyways. In my mind there are few ways a hash can error out

mscrypto-bcrypt/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+impl BcryptProviderBuilder {
+    /// Records an algorithm that `build()` must confirm is supported.
+    pub fn require(mut self, algorithm: Algorithm) -> Self {
+        self.required.push(algorithm);
+        self
+    }
+
+    /// Probes SHA-3 availability, then verifies the require-list. SHA-2 needs no
+    /// setup (pseudo-handles), so it never fails here.
+    pub fn build(self) -> Result<BcryptProvider, ProviderBuildError> {
```
nnmkhang, line 79: This is where we would error if SHA3 is not available

mscrypto-bcrypt/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+                Sha3Algorithm::Sha3_256 => self.sha3_256,
+                Sha3Algorithm::Sha3_384 => self.sha3_384,
+                Sha3Algorithm::Sha3_512 => self.sha3_512,
+            },
+            _ => false,
+        }
+    }
+}
+
+fn backend_info() -> BackendInfo {
```
nnmkhang, line 126: will try to find a way to use bcrypt / windows version number as a pseudo version

nnmkhang: Now looking at it, might make sense for the version to just be a string, then we can do checks per provider rather than forcing bcrypt into the symcrypt semver semantics.

mscrypto-symcrypt/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+}
+
+// Single point where SymCrypt module usability is verified when a provider is
+// built. SymCrypt checks version compatibility during its lazy initialization on
+// first use, which aborts on a mismatch, so there is no recoverable failure to
+// report today. A graceful module-init entry point will surface an incompatible
+// module here as `ProviderBuildError::Backend { backend, operation }`.
+fn initialize_module() -> Result<(), ProviderBuildError> {
```
nnmkhang, line 115: will replace with `SymCryptModuleInitEX` so we can probe during build and graceful error on old versions of SymCrypt.

nnmkhang: We will only support new versions of symcrypt.dll for this crate

mscrypto-symcrypt/build.rs
```diff
@@ -0,0 +1,19 @@
+// Records how symcrypt-sys is linked so the provider can report it through
```
nnmkhang, line 1: WIP

mscrypto/src/hash.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/// Provider-facing construction seam. Not part of the caller-facing API: provider
+/// crates import `BuildDigest` to produce a `Digest` from a backend's output. It
+/// lives here rather than as an inherent method so it stays off `Digest`'s public
+/// surface; it is reachable cross-crate only because Rust has no friend-crate
+/// visibility. Ordinary callers receive a `Digest` from `Hash::digest` /
+/// `HashOps::finalize` and read it via `as_bytes` / `as_ref`.
+#[doc(hidden)]
+pub mod provider {
```
nnmkhang, line 49: This is separate since every API needs to be public for the trait to work. However, this is an implementation detail for the provider and should not be exposed to the caller. We can't physically stop them from using this. but hiding it like this from the docs is the best work around imo.

PR #121, opened 2026-09-10
 add MLKEM

nnmkhang:
### Description of Changes:
Add MLKEM

### Breaking Changes if any:

None

### ✅ Admin Checklist
- [x] PR targets the active `release/X.Y.Z` branch.
- [x] Review the PR description and ensure all necessary details are included.
- [x] Update/add unit tests for changed code.
- [x] Update/add documentation for new or changed APIs.
- [x] Run `cargo test --all-features` on Windows and WSL.

Check the guidelines in [`CONTRIBUTING.md`](../CONTRIBUTING.md) for more info.


nnmkhang:
@copilot+gpt-6-astra review the code

rust-symcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,872 @@
...
+            }
+            MlKemParams::MlKem1024 => {
+                symcrypt_sys::_SYMCRYPT_MLKEM_PARAMS_SYMCRYPT_MLKEM_PARAMS_MLKEM1024
+            }
+        }
+    }
+}
+
+// Rust only exposes parameter and format values accepted by the size-query APIs.
+fn unreachable_size_error(function: &str, err: symcrypt_sys::SYMCRYPT_ERROR) -> ! {
```
mlindgren, line 128: The `!` type is a newly stablizied feature, right? I just saw a video about it recently. Cool to see it being used already

rust-symcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,872 @@
...
+        self.key_format_len(MlKemKeyFormat::EncapsulationKey)
+    }
+
+    /// Returns the decapsulation key size in bytes.
+    pub fn decapsulation_key_len(&self) -> usize {
+        self.key_format_len(MlKemKeyFormat::DecapsulationKey)
+    }
+
+    /// Returns the ciphertext size in bytes.
+    pub fn ciphertext_len(&self) -> usize {
```
mlindgren, line 80: I'm of two minds about this and `key_format_len`. On the one hand, I like keeping as much logic in SymCrypt as possible, so that the Rust layer is just a thing wrapper and we don't have to update in two places if more sizes are standardized in the future, and I think that has been the guidance I've given in the past. On the other hand, calling into an FFI function for what is effectively a single match statement kind of feels like overkill. @samuel-lee-msft what do you think?

rust-symcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,872 @@
...
+                symcrypt_sys::SymCryptMlKemkeyFree(self.0);
+            }
+        }
+    }
+}
+
+/// A 32-byte ML-KEM shared secret.
+///
+/// The bytes are wiped on drop and omitted from `Debug`.
+pub struct SharedSecret([u8; SHARED_SECRET_LEN]);
```
mlindgren, line 177: Not blocking, but we should probably have a secret buffer type representing an arbitrary-length array that is wiped on drop

rust-symcrypt/src/mlkem.rs
```diff
@@ -0,0 +1,872 @@
...
+
+/// An ML-KEM key.
+///
+/// A key may contain a full key pair or only an encapsulation key.
+#[derive(Debug)]
+pub struct MlKemKey {
+    inner_key: InnerMlKemKey,
+    params: MlKemParams,
+    has_private_key: bool,
+    has_private_seed: bool,
```
mlindgren: I believe these are already tracked by SymCrypt. Do we need to track them here as well? Duplicating state across related objects always risks desynchronization

mlindgren (changes requested):
lgtm except that I'd prefer not tracking the private key/private seed state of the MlKemKey in this layer unless there's a strong reason for it

rust-symcrypt/src/mlkem.rs
samuel-lee-msft: Yeeah, I think it's probably better to layer it like this unless/until we provide the Rust interface directly from SymCRust code. We can always reevaluate if we see any real world drawback (perf? versioning?) to the additional FFI.

mlindgren: I'm cool with that. Marking as resolved.

mlindgren:
@copilot Your review comment stated "The security-sensitive FFI and concurrency guarantees warrant final human review." Can you expand on what you mean by this? The types in the crate do not implement Send or Sync and thus cannot be transferred or shared between threads, by design (for now anyway).

rust-symcrypt/src/lib.rs
```diff
@@ -9,6 +9,7 @@ pub mod gcm;
 pub mod hash;
 pub mod hkdf;
 pub mod hmac;
+pub mod mlkem;
```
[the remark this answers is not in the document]
nnmkhang: Fixed

rust-symcrypt/src/mlkem.rs
nnmkhang: Removed this and the MLKEM format, this info is inferred by calling symcrypt api. Added a comment to explain

rust-symcrypt/src/mlkem.rs
nnmkhang: This pattern is the same across the other places in the code base. it should be wiped via the inner key when it goes out of scope

mlindgren:
Ah, I missed the Send/Sync impls for MlKemKey, my bad. But the comment is correct that the key object is safe for concurrent read-only use, and the object is immutable after construction.

mlindgren (approved):
lgtm

rust-symcrypt/src/mlkem.rs
nnmkhang: yeah pretty cool stuff

